PR #117, opened 2015-01-21
package removal from list

jpoirier:


dom96:
Why are you removing these packages?

PR #128, opened 2015-02-15
nimrat added

apense:
it's a rational arithmetic module

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -1236,5 +1236,14 @@
     "description": "Module for working with paths that is as similar as possible to Python's pathlib",
     "license": "MIT",
     "web": "https://github.com/achesak/nim-pythonpathlib"
+  },
+  {
+    "name": "nimrat",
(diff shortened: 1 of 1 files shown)


dom96:
I don't think https://github.com/apense/nimrat/blob/master/nimrat.nimble#L8 is required in your .nimble file.

apense:
Agreed. Just removed it.

apense:
But, you should reject this, I think. I just noticed that a rationals library was added to the standard lib.

dom96:
Not necessarily. But it's your choice.

apense:
I reopened it. Sorry for the indecision. This will hopefully be useful until the better module is added. But at least I can use this to write a number theory module for now.

dom96:
I don't mind libraries which serve the same purpose as ones already in the standard library being added to this packages repo. Same goes for libraries which are already in this repo.

PR #135, opened 2015-03-01
A simple package scanner to check packages.json

FedericoCeratto:
Requires a Git API token and the Python Requests, Setproctitle and PyGithub libraries

(I know, it's in Python... * ducks *)

### DIFF
--- a/nim_package_scanner.py
+++ b/nim_package_scanner.py
@@ -1,0 +1,138 @@
+#!/usr/bin/env python
+
+"""
+A very simple Nim package scanner.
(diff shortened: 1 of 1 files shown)


flaviut:
cc @dom96 @Varriount 

> (I know, it's in Python... \*ducks\*)

This doesn't do anything that isn't doable out-of-the box with Nim, even the github module is unnecessary. It should be simple enough to port it to Nim.

dom96:
Port it to Nim please.

PR #136, opened 2015-03-02
Add msgpack

akiradeveloper:
Signed-off-by: Akira Hayakawa <ruby.wktk@gmail.com>

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -1281,5 +1281,14 @@
    "description":"A component/entity system",
    "license":"CC0",
    "web":"https://bitbucket.org/fowlmouth/entoody"
+  },
+  {
+    "name": "msgpack",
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -1281,5 +1281,14 @@
    "description":"A component/entity system",
    "license":"CC0",
    "web":"https://bitbucket.org/fowlmouth/entoody"
+  },
+  {
+    "name": "msgpack",
+    "url": "https://github.com/akiradeveloper/msgpack-nim",
```
fowlmouth: This url should end with `.git` or start with `git://`

akiradeveloper:
Thanks for reviewing. So my fuse library needs to append ".git" too. I will fix it later.

akiradeveloper:
pull please? (still mistake? or do you think not worth?)

PR #137, opened 2015-03-02
Fixed NimQML renaming to nimqml

filcuc:
I renamed the NimQML to nimqml


dom96:
package names are case insensitive but i'll merge anyway

filcuc:
Ah ok didn't knew that ;)

PR #138, opened 2015-03-07
Add package_scanner.nim

FedericoCeratto:
...this time it is implemented in Nim. It requires no GitHub API token as it simply check the reachability of the source code repository.
Every error except unexpected license names is counted and returned as exit value.

### DIFF
--- a/package_scanner.nim
+++ b/package_scanner.nim
@@ -1,0 +1,101 @@
+
+# A very simple Nim package scanner.
+#
+# Scan the package list from:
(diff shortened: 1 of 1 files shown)


dom96:
I'd have to run this every single time manually though... I'd prefer an automatic solution.

FedericoCeratto:
TravisCI does the trick - see an example at https://travis-ci.org/FedericoCeratto/packages/jobs/53537960

FedericoCeratto:
See #139

PR #139, opened 2015-03-10
Related to  #138, add package scanner and .travis.yml

FedericoCeratto:


flaviut:
LGTM

/cc @dom96 Lets get this merged?

dom96:
I would need to set up travis for this I guess?

FedericoCeratto:
@dom96: enable TravisCI on GitHub? Yes.
Setting up a local Travis instance on your dev host? No

fowlmouth:
@dom96 is there a travis-ci account for nim-lang if not I can set up one up.

.travis.yml
```diff
@@ -0,0 +1,23 @@
...
+os:
+  - linux
+
+language: c
+
+env:
+  - BRANCH="travis"
+  - BRANCH="devel"
+
+install:
```
megawac: Probably should be `before_install`; `install` would generally cover things like `nimble install` or `nimble build`

.travis.yml
```diff
@@ -0,0 +1,23 @@
+
+os:
+  - linux
+
+language: c
+
+env:
+  - BRANCH="travis"
+  - BRANCH="devel"
```
megawac: Why do you want to setup a matrix?

package_scanner.nim
```diff
@@ -0,0 +1,101 @@
...
+
+    elif not (pdata["method"].str in VCS_TYPES):
+      echo "E: ", name, "has an unknown method: ", pdata["method"].str
+      result.inc()
+
+    if not pdata.hasKey("license"):
+      echo "E: ", name, "has no license"
+      result.inc()
+    elif not (pdata["license"].str in LICENSES):
+      echo "W: ", name, "has an unexpected license: ", pdata["license"]
```
megawac: Is there a good reason to be so strict with licencing as to fail a build if it doesn't appear in that list? For example you're missing `MPL` which I'd say is entirely valid. Also is case sensitivity important in a license?

package_scanner.nim
```diff
@@ -0,0 +1,101 @@
...
+
+      except HttpRequestError, TimeoutError, AssertionError:
+        echo "E: ", name, ": unable to fetch repository ", url, " ", getCurrentExceptionMsg()
+        result.inc()
+
+  echo "Error count: ", result
+  return
+
+when isMainModule:
+  quit(check())
```
megawac: Will this generate the right exit code? I think travis will only fail for `0` responses

flaviut: The exact opposite actually:
>All commands must exit with code 0 on success. Anything else is considered failure.

megawac: haha I just realized that

package_scanner.nim
flaviut: The array might be incomplete, but having canonical names for licenses is important for tooling.

PR #145, opened 2015-04-12
Beanstalkd

tormaroe:
Would love to have beastalkd.nim included in the nimble index. If you want to provide any feedback on the module as well (I'm learning the language), that would be a great bonus :)

Btw, I tried running `nible build` in my root, but nothing happened. I can find no documentation on this, so I basically haven't verified the .nimble file. Should be ok though.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -1326,5 +1326,15 @@
     "description": "Bindings for Closure Compiler web API.",
     "license": "MIT",
     "web": "https://github.com/yglukhov/closure_compiler"
+  },
+  {
+    "name": "beanstalkd",
+    "url": "git@github.com:tormaroe/beanstalkd.nim.git",
+    "method": "git",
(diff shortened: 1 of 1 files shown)


dom96:
You shouldn't specify the SSH git url as I think it will only work for people who have write access to your repo.

dom96:
Nothing happened when you executed ``nimble build`` because your nimble file doesn't specify any modules that need to be built. You do so by writing ``bin = "file"``

tormaroe:
Git url fixed in new commit. I thought `nimble build` was some way to verify the package spec, but I realize I was wrong. Thank you.

dom96:
If you want to verify it then run ``nimble install`` :)

tormaroe:
Ahhhh :) Tried it, seems to have installed it properly. Thanks again

dom96:
np :)

PR #154, opened 2015-05-12
Added nshout - libshout bindings

Senketsu:
Personal libshout wrapper used in one of my projects.
Someone might find it usefull


packages.json
```diff
@@ -1426,4 +1426,13 @@
     "license": "MIT",
     "web": "https://github.com/jangko/nimpdf"
   }
```
dom96: Missing a comma here.

PR #156, opened 2015-05-14
Added library nuuid

wheineman:
Added a Nim source only UUID generator library


dom96:
These are incorrect in the .nimble file: https://github.com/wheineman/nim-only-uuid/blob/master/nuuid.nimble#L7,L9

I will merge this in anyway, but you should fix it.

wheineman:
Thanks, fixed. I'll get the hang of this.

PR #162, opened 2015-05-17
Removes dead ouroboros package.

gradha:


dom96:
Aww. Why did you remove it?

PR #166, opened 2015-05-20
add msgpack4nim package

jangko:
msgpack4nim is another MessagePack implementation written in pure nim


dom96:
I'm curious, why another one?

jangko:
@dom96 thanks for your curiosity, nim is a great language, and a great language needs a great ecosystem

PR #169, opened 2015-05-21
improved(?) package_scanner!

fenekku:
- package_scanner.nim was not checking for "tags" even though "tags" are required per https://github.com/nim-lang/packages#adding-your-own-package so that was fixed
- A package with a missing name and other errors counted as 1 error,
  while a package with a missing description and a missing method for instance
  counted as 2 errors. Strange. There are two possible approaches: count the total number of errors
  or count the total number of erroneous packages. "Why count at all?" one could also argue :grin: .
  I decided to count the number of erroneous packages.
- Refactored the checks for clarity of control flow
- :warning: `package_scanner` still does the "wrong" thing of checking for `web` url and not the real `url`. I have conceptual ideas, but I don't know how to implement them...yet!
- Other :warning: Fetching url of all packages when only one (assumed most common case scenario) changes does not scale super well :innocent:

### DIFF
--- a/package_scanner.nim
+++ b/package_scanner.nim
@@ -4,11 +4,12 @@
 # Scans the package list from this repository.
 #
 # Check the packages for:
-#  * Missing/unknown license
-#  * Missing description
 #  * Missing name
 #  * Missing/unknown method
 #  * Missing/unreachable repository
+#  * Missing tags
+#  * Missing description
+#  * Missing/unknown license
 #
 # Usage: nim c -d:ssl -r package_scanner.nim
 #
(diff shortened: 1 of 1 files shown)


package_scanner.nim
```diff
@@ -41,24 +42,44 @@ const
...
+  # TODO: Make this check the actual repo url and check if there is a
+  #       nimble file in it
+  result = true
+  var url: string
+
+  if not urlJson.isNil:
+    url = urlJson.str
+
+    try:
+      discard getContent(url, timeout=1000)
```
dom96: Won't this fail for ``git://`` URLs?

dom96:
What would be ideal is if we could test installing all the nimble packages (using nimble), that may take too much time though.

package_scanner.nim
fenekku: As mentioned, we were not actually using the git:// URLs before and we are not now either. Notice how it is the "web" URL that is used and not the "url" URL that is used. :wink: 

The TODO is about fixing that but fixing that is hard. See my reply below.

fenekku:
So I was thinking about this quite a lot these last days.  :pensive: 

Checking or installing all packages when, most common case scenario, only one changes does not scale well at all. Ideally, only the changed packages are checked. I didn't know how to do this until I wrote my thought process, but I think I have a proposition now. Let me share it with you :thought_balloon: 

My thought process went along these lines.

Force the process of package upload to be made through nimble and have nimble ensure all these things are good locally. I don't know how to do this / lots of work maybe / research into how others do it / need for systemic change.

Alternative: Have a cache of the package list that is checked against the newly pushed package list. New or modified packages are detected and only those are checked. On Travis caching is a [paid option](http://docs.travis-ci.com/user/caching/). On CircleCi it is not: https://circleci.com/docs/how-cache-works. That cache should be created by package_scanner.nim (it will not be part of the code repository). This is feasible (I think :sweat_smile:) and will allow us not to take too much time thanks to the caching.

Now we need to figure out if a package has a .nimble file (and warn if a .babel file is found). Thought process:

  * [`git archive`](http://git-scm.com/docs/git-archive) would have been a nice tool to get the top level repository content of the package. We could then look into whether the `.nimble` file is there. However GitHub does not allow `git archive` http://stackoverflow.com/questions/1178389/browse-and-display-files-in-a-git-repo-without-cloning/1178586#1178586.
  * We could scrape the HTML page of the package but that means keeping up with the front-ends which can be anything. Nope.  
  * That leaves us with cloning the repos. This will take a first time hit, but then the caching should help us out. We can even make sure that the cloning is shallow.

I think we have a potential approach here. Clone and cache. I could work on it but it would mean replacing travis with circleci. Also I have not thought about how to do cache invalidation yet...
 
Does that make sense?

dom96:
I think we'll be fine with what we have so far. :)

PR #178, opened 2015-05-28
Removing trailing comma

Nycto:
It is breaking the Nim website under the package list:
http://nim-lang.org/docs/lib.html

```
Unable to retrieve package list: Unexpected token ]
```


dom96:
Sorry, didn't realise you made a PR for this. I did myself.

PR #181, opened 2015-06-01
libs renamed; libovr added

bluenote10:
renamed nim-heap to binaryheap and nim-stringinterpolation to stringinterpolation


dom96:
thanks

PR #195, opened 2015-06-21
added redis, niminst, dialogs packages

Araq:


dom96:
None of these have ``.nimble`` files in their repo...

PR #214, opened 2015-07-21
Add https://github.com/stavenko/nim-glm/

stavenko:


dom96:
Don't name packages with dashes please. Just remove the "nim-" prefix.

stavenko:
Done. Seems to be working.

dom96:
Thanks.

PR #242, opened 2015-08-24
Remove git conflict marker from packages.json

andrew:
Makes it valid JSON again so that https://libraries.io can parse it correctly :panda_face:


dom96:
Thanks for the PR. I didn't notice it and just fixed it myself heh.

PR #245, opened 2015-08-25
Add POP3 client library

FedericoCeratto:


dom96:
This package's name should match the name in your nimble file.

PR #247, opened 2015-08-27
Add project nim-rpc

rogercloud:


rogercloud:
Fix error on travis-ci "E: lapp: unable to fetch repository https://gitlab.3dicc.com/gokr/lapp Call to 'readLine' timed out."

dom96:
Many issues with your package:

* There should be no dashes in package names.
* You need to put the modules your package exposes in a nimrpc or rpc or pkgname directory. Otherwise your modules will pollute the global namespace.
* You should also move your modules into a ``src`` directory and then use ``srcDir = "src"`` in your nimble file, so that the examples are not installed.
* Add your package listing at the bottom of packages.json not in the middle of it.

The 3dicc gitlab issues are intermittent. I don't think adding ``.git`` will help.

rogercloud:
Thank you for the advice. Issues fixed and revert lapp fix.

PR #253, opened 2015-09-16
added 'nim-db-ex'

Senketsu:
extension modules for Nim's 'db_*' modules


dom96:
Your package name should not contain dashes. The modules that your package exposes need to be namespaced under a directory named after your package.

PR #258, opened 2015-09-21
Add playlists.

achesak:


dom96:
Nice! This will really come in handy.

PR #259, opened 2015-09-21
fix: LiteStore -> litestore

h3rald:
The name of the package should match the name of the compiled binary file really...


dom96:
Pretty sure it's case insensitive anyway :)

PR #302, opened 2016-01-07
Added TeaFiles and Emmy

andreaferretti:
Added TeaFiles and Emmy


dom96:
Awesome to see Nim repos in the @unicredit org :)

andreaferretti:
Not the [first one](https://github.com/unicredit/csvtools), [either](https://github.com/unicredit/linear-algebra) ;-)

PR #303, opened 2016-01-08
Renamed nim-libclang to libclang.

yglukhov:
Fixes https://github.com/jovial/nim-libclang/issues/2
@jovial, Please acknowledge.


dom96:
For these renames, please add an alias (i.e. another package listing underneath the one you want to rename). That way old packages depending on these can still work.

yglukhov:
New pr: #304

PR #311, opened 2016-01-30
Renamed packages to follow upcoming nimble package naming conventions.

gmpreussner:


dom96:
Thanks! But could you instead add new entries for these at the bottom of the file? That way the old names will continue to work for people who have used them already. :)

gmpreussner:
Oh yes, didn't think of that :)

gmpreussner:
Should be fixed now. We might want to add some field for deprecated/obsolete packages, so that they don't show up in new queries.

dom96:
Agreed :)

dom96:
Thank you for fixing!

PR #314, opened 2016-02-13
Add nimbench package

ivankoster:
nimbench is supported on platforms windows, macos and posix.
Request to add nimbench 0.1.1 to nimble package listing.


dom96:
Nice. Welcome to Nim land :)

PR #317, opened 2016-02-15
chipmunk update

matkuki:


dom96:
Please add another entry with the new name and leave the old entry in there.

matkuki:
I'm also updating BlaXpirit's ```chipmunk7``` wrapper, which will be added to the packages. 
So it seems more appropriate to have modules '```chipmunk6``` and ```chipmunk7```' instead of
'```chipmunk``` and ```chipmunk7```'. The first option looks clearer to me.
I'm fine either way, but are you sure?

dom96:
You should do the following then: ``chipmunk``, ``chipmunk6`` and ``chipmunk7``, where ``chipmunk`` and ``chipmunk6`` point to the same package.

PR #319, opened 2016-02-19
Remove trailing slashes from url/web links for boydgreenfield/onecode…

boydgreenfield:
…x packages

I'm investigating how often this causes failures, but we're seeing issues w/ older versions of git and this syntax. Once I nail it down entirely, happy to open a PR that does this _across_ all the repos in packages.json.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -762,7 +762,7 @@
   },
   {
     "name": "bloom",
-    "url": "git://github.com/boydgreenfield/nimrod-bloom/",
+    "url": "git://github.com/boydgreenfield/nimrod-bloom",
     "method": "git",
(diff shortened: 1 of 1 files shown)


dom96:
Interesting issue. It might be better long-term to fix this in Nimble, i.e. have it remove the trailing slashes. I will merge this in the meantime. Could you create an issue in the Nimble repo for this?

dom96:
Looks like others are having the same issue. No need to report this separately, #204 tracks it.

boydgreenfield:
Ok, thanks @dom96! – and sorry for the delay. Got caught up for a few days there.

PR #323, opened 2016-03-05
added sdl2-nim

Vladar4:


trustable-code:
I'm just curious: What is the difference to https://github.com/nim-lang/sdl2 ?

Vladar4:
Hello.

Surely, two wrappers for one library can't be much different, but 
besides some implementation differences, sd2-nim have fully formatted 
documentation in Nim style and bunch of practical examples, which helped 
me in debugging. Also original file structure and functions' names 
preserved (except for SDL_ prefix, of course).

PR #326, opened 2016-03-20
Added Rosencrantz

andreaferretti:
Added Rosencrantz

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -4370,5 +4370,19 @@
     "description": "Library for shell scripting in nim",
     "license": "MIT",
     "web": "https://github.com/vegansk/nimshell"
+  },
+  {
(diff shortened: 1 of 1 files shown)


dom96:
Looks awesome. But you will need to rename ``rosen`` to ``rosencratz`` to satisfy Nimble's package guidelines. (or you can rename the package name to ``rosen`` instead if you want).

dom96:
See here for details about the guidelines/conventions: https://github.com/nim-lang/nimble#libraries

These will become enforced by Nimble in a future update.

andreaferretti:
Fixed! For some reason, I believed that having both a file `rosencrantz.nim` and modules in the `rosencrantz` directory would lead to conflicts

PR #332, opened 2016-03-31
Add package xxtea_nim

andot:


dom96:
Please rename the package to ``xxtea``. Your [module](https://github.com/xxtea/xxtea-nim/blob/master/xxtea.nim) name must match the package name.

PR #337, opened 2016-04-05
Add asyncdocker

tulayang:
This module implements an Docker Engine client based on Docker Remotet API. It's
asynchronous (non-blocking) that it can be used to write web services for deploying
swarm cluster and containers automatically on cloud environment. Of course, it
can also be used to write any local deployment tools.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -4510,5 +4510,17 @@
     "description": "Additions to the Nim's standard library, like boost for C++",
     "license": "MIT",
     "web": "https://github.com/vegansk/nimboost"
+  },
+  {
+    "name": "asyncdocker",
+    "url": "https://github.com/tulayang/asyncdocker",
+    "method": "git",
+    "tags": [
+      "async",
+      "docker"
+    ],
+    "description": "Asynchronous docker client written by Nim-lang",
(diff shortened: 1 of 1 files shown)


dom96:
Beautiful! Looks like I need to learn Docker :)

What's the reasoning behind ``asynchttpclient``? Was ``httpclient`` not sufficient? Maybe you could create some PRs?

tulayang:
@dom96

Thanks!

I noticed ``httpclient`` provide  asynchronous request, but I also noticed "As such in the future this module may change, and the current procedures will be deprecated.", so I wrote a separate implementation. 

In addition, some REST API of Docker Remote API does not conform to the standard HTTP 1.1, such as ``Content-Type: application/vnd.docker.raw-stream`` --- no ``Content-Length``, no ``Transfer-Encoding: chunked``, no ``Connection: close``, but the scheme is ``HTTP 1.1``!  This is contrary to the standard HTTP 1.1. 

So,  I need some similar plug-in functions if I used ``httpclient`` to extend it. 

However, much of ``asynchttpclient`` are taken from ``httpclient``. Thanks for the contributors of ``httpclient``. :)

PR #339, opened 2016-04-06
Add classifiers

FedericoCeratto:
Based on the Trove classifiers, to be used as tags for Nimble packages

### DIFF
--- a/classifiers.txt
+++ b/classifiers.txt
@@ -1,0 +1,467 @@
+Development Status :: 1 - Planning
+Development Status :: 2 - Pre-Alpha
+Development Status :: 3 - Alpha
+Development Status :: 4 - Beta
+Development Status :: 5 - Production/Stable
+Development Status :: 6 - Mature
+Development Status :: 7 - Inactive
+Environment :: Console
+Environment :: Console :: Curses
+Environment :: Console :: Framebuffer
+Environment :: Console :: Newt
+Environment :: Console :: svgalib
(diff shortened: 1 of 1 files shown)


dom96:
I don't think our tags need a hierarchy. Please look at the tags we currently have in the repo and create a list of them, and add some other sensible ones.

FedericoCeratto:
> I don't think our tags need a hierarchy.
Why?
Btw, I ran a statistic on the tags. Being just words, not all of them are meaningful or searchable. The Trove classifiers are meant to be more precise. Also, they would allow to match supported Nim versions and subversions, e.g. "Programming Language :: Nim :: 1." would match 1.*

dom96:
That ambiguity can be easily resolved with the addition of a "library" tag (which many packages already employ). Tags are much easier to search for than a long hierarchical category.

alex-laskin:
Today I was trying to choose good tags for my library, so I took all tags from packages.json, sorted them and select only unique ones. And there are many tags like "2D" and "2d", "MPD" and "Music Player Daemon", "algorithm" and "algorithms", "app" and "application", and I can list them all day long. 659 (not so) unique tags. And I hope, that at avg 4 tag per package, this list will became a total mess pretty quickly.

dom96:
I'm going to politely close this PR because I still haven't changed my mind about this.

We should standardise some tags, but I don't think we need to limit which tags are allowed. As an example, look at GitHub's new tag features.

PR #342, opened 2016-04-12
Add package pledge

euantorano:
Adding the `pledge` package, which adds an easy accessor to OpenBSD's [`pledge(2)`](http://man.openbsd.org/OpenBSD-current/man2/pledge.2). On unsupported operating systems (e.g.: not OpenBSD), it provides a replacement function that always succeeds.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -4547,5 +4547,17 @@
     "description": "json de/serializer for tuples and more",
     "license": "MIT",
     "web": "https://github.com/niv/jser.nim"
+  },
+  {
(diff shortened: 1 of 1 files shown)


dom96:
Nice! Would be cool to see an enum instead of a string as a parameter to `pledge` :)

euantorano:
@dom96 Yep, that's the plan. Just wanted to get a version working first :smile:

dom96:
Awesome, just saw your commit :D

PR #346, opened 2016-04-23
Adding "cucumber_nim"

shaunc:
As a "learning nim" project I've created a port of the cucumber testing framework into nim. As this is my first project, I'm sure it could use a good deal of cleanup. It does run its own tests, at least.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -4601,5 +4601,16 @@
     "description": "websockets for nim",
     "license": "MIT",
     "web": "https://github.com/niv/websocket.nim"
+  },
+  {
+    "name": "cucumber",
+    "url": "https://github.com/shaunc/cucumber_nim",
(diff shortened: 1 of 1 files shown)


dom96:
Nice, you will need to rename https://github.com/shaunc/cucumber_nim/blob/master/cucumber_nim.nimble to ``cucumber.nimble`` though.

shaunc:
ah ... and rename cucumber/ to cucumber.pkg/ I guess...?

shaunc:
hmm... using "cucumber.pkg" doesn't seem to work:

    Error: cannot open 'cucumber/types'

I guess "cucumber.nim" isn't a main program... just a convenience module for imports

dom96:
Nope. It should be named ``cucumber/``.

PR #348, opened 2016-04-28
Nim for awk programmers

greencardamom:
Library for porting awk to nim

Apparently I did something wrong but I don't know what.

Update: forgot a comma in the json file


dom96:
Cool package! Thanks :)

PR #351, opened 2016-05-12
Added sph

aidansteele:
This is my first Nimble package, so please let me know what I did wrong - I'm sure I made a mistake or six. :)


dom96:
Looks good except one thing:

* Your `sphmacros` module should be moved into `sph/private/sphmacros` if you don't want user's of your package to import it, if you do then into `sph/sphmacros`

Also I'm not sure how well nimble handles git submodules. I guess we'll see :)

aidansteele:
Thanks for the prompt merging. I'll make that change now, thanks for the advice! 😄

PR #353, opened 2016-05-14
Add libsodium wrapper

FedericoCeratto:


dom96:
you will need to put ``sodium_sizes`` into a `libsodium` directory (Nimble will enforce this in the future)

PR #355, opened 2016-05-18
Added persistent_enums

yglukhov:


yglukhov:
@dom96, it looks like there was some connection problem during the test:
```
E: grp: unable to fetch repository https://github.com/achesak/nim-grp Call to 'readLine' timed out.
```
Could you restart them please?

dom96:
Guess the tester needs to be less strict with timeouts.

PR #359, opened 2016-06-02
Replaced old yaml package with NimYAML

flyx:
Author of the old package agreed to remove it from `packages.json` and replace it with NimYAML:

https://github.com/nim-appkit/yaml/issues/1


dom96:
please keep ``nimyaml`` pointing to your package, that way people who have it in their nimble file as a dependency can still use it.

PR #363, opened 2016-06-09
Use secure https instead of insecure git protocol

def-:
git protocol is absolutely insecure, so we should not be using it by default. If a user really wants to use `git://` they can set this in their git config:

    [url "git://"]
        insteadOf = https://

### DIFF
--- a/package_scanner.nim
+++ b/package_scanner.nim
@@ -10,6 +10,7 @@
 #  * Missing tags
 #  * Missing description
 #  * Missing/unknown license
+#  * Insecure git:// url on GitHub
 #
 # Usage: nim c -d:ssl -r package_scanner.nim
 #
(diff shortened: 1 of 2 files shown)


dom96:
This is already covered by Nimble: https://github.com/nim-lang/nimble/blob/master/src/nimble.nim#L495

But I'm happy to merge this anyway in case other tools will misuse the ``git://`` protocol.

package_scanner.nim
```diff
@@ -116,6 +117,10 @@ proc check(): int =
       echo "E: ", name, " has no license"
       result.inc()
 
+    elif pdata["url"].str.normalize.startsWith("git://github.com/"):
```
dom96: What about BitBucket, GitLab, etc? This check shouldn't be GitHub-specific, should it?

def-: I have no idea which hosters support https and if they use the same URL. I know that it would be problematic for self-hosted repositories, haven't used any of the others.

dom96: Good point.

PR #368, opened 2016-06-22
Add `semver` package

euantorano:
Add the semver package, this time manually rather than using `nimble publish`.


dom96:
btw you might be interested in Nimble's ``version`` module: https://github.com/nim-lang/nimble/blob/master/src/nimblepkg/version.nim

You can import it via ``import nimblepkg/version``.

euantorano:
Interesting. I assumed such a thing must exist. I might steal a couple of prices from there.

PR #369, opened 2016-06-24
Add package ad

subsetpark:


dom96:
Please don't use ``nimble publish`` until https://github.com/nim-lang/nimble/issues/241 is fixed.

PR #375, opened 2016-07-04
Added luna Lua scripting library

smallfx:


packages.json
```diff
@@ -4889,5 +4889,17 @@
     "web": "https://github.com/miere43/nim-registry",
     "license": "MIT",
     "method": "git"
+  },
+  {
+	"name": "luna.nim",
```
dom96: Spaces instead of tabs please :)

smallfx: @dom96 my bad! Fixed now.

packages.json
```diff
@@ -4889,5 +4889,17 @@
     "web": "https://github.com/miere43/nim-registry",
     "license": "MIT",
     "method": "git"
+  },
+  {
+    "name": "luna.nim",
```
dom96: Sorry, one more thing. You should just call your package ``luna``. Rename your ``.nimble`` file to ``luna.nimble`` as well and make sure the ``name`` field inside is set to ``luna``.

dom96: The name of your github repo doesn't need to be renamed.

smallfx: @dom96 changed it :)

dom96:
Thanks!

PR #378, opened 2016-07-09
Add package circleci_client

FedericoCeratto:


dom96:
It seems that ``nimble publish`` is now reordering all the package's fields? 😠 *sigh*

PR #383, opened 2016-07-20
Add package autome

miere43:


miere43:
well I'll do that with hands

PR #384, opened 2016-07-20
Add autome library

miere43:


dom96:
Thank you for re-creating manually, ``nimble publish`` is a bit broken.

PR #392, opened 2016-08-03
Add package foliant

moigagoo:


dom96:
Please don't use ``nimble publish`` until it is fixed.

moigagoo:
Sorry! Just to clarify, until what is fixed? Also, to update a package, do I use `nimble publish`?

dom96:
This one: https://github.com/nim-lang/nimble/issues/241

Depends what you want to update. But generally, you shouldn't be using ``nimble publish`` at all until that issue above is fixed.

moigagoo:
@dom96 Thanks! I wasn't aware `nimble publish` didn't work. I thought my package was being premoderated or something like that.

> Depends what you want to update.

If I'm releasing a new version of the package, should I use `nimble publish`? When it's back on track, of course.

dom96:
> @dom96 Thanks! I wasn't aware nimble publish didn't work. I thought my package was being premoderated or something like that.

It works but it's creating huge diffs so it's buggy.

> If I'm releasing a new version of the package, should I use nimble publish? When it's back on track, of course.

Nope. Versions are all stored via Git Tags in your repo so all you have to do is update your Git repo. The packages repository only stores URLs to different package's repos.

PR #408, opened 2016-08-31
JSON does not have an escape sequence for '

flyx:
Nim's `json` module allows it, but the native parser used for JS does not, leading to this error:

http://forum.nim-lang.org/t/2504

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -5192,7 +5192,7 @@
     ],
     "method": "git",
     "license": "MIT",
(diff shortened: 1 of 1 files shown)


dom96:
JS' native parser is sooo annoying.

dom96:
Thanks for the fix.

flyx:
I wonder if it is a good idea to support `\'`in the Nim JSON implementation when JS does not support it. That leads to inconsistent behaviour across targets.

PR #413, opened 2016-09-16
Added SDF.nim

Halsys:


packages.json
```diff
@@ -5264,5 +5264,22 @@
     "description": "ISAAC PRNG implementation on Nim",
     "license": "MIT",
     "web": "https://github.com/pragmagic/isaac/"
+  },
+  {
+    "name": "SDF.nim",
```
dom96: Remove the ``.nim``, the name needs to match your package name.

Halsys: I made the change but just a small question. Would `SDF.nim.nimble` be an also acceptable package filename? Sorry.

Halsys: Or does this have a similar effect like dashes in names?

dom96: I think it does. You would then need to rename your module to ``SDF.nim.nim`` and import it via ``import SDF.nim``, I dunno, maybe that works. Best way to check is to test it.

I'll merge this for now.

PR #415, opened 2016-09-23
Added chipmunk7_demos package

matkuki:


dom96:
Why doesn't this package depend on a ``chipmunk`` package?

matkuki:
A slighty modified version of ```chipmunk``` is embedded in it.

matkuki:
Keep in mind that this is for the Chipmunk7 library, which BlaXpirit has yet to add to nimble!
Also a reason why I embedded it.

PR #423, opened 2016-11-10
Rename geoip

achesak:
As to reduce confusion with mmgeoip.


dom96:
When renaming, please duplicate the entries and keep the old name for one of them.

achesak:
Added as new entry

PR #425, opened 2016-11-21
Added nim bind for forestDB.

rivasiv:
Exdend with forestDB. Base is under https://github.com/couchbase/forestdb. Nim bind is under: https://github.com/nimscale/forestdb.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -5450,5 +5450,20 @@
     "description": "Bindings for libjwt",
     "license": "LGPLv2.1",
     "web": "https://github.com/nimscale/nim-libjwt"
+  },
+  {
+    "name": "forestdb",
+    "url": "https://github.com/nimscale/forestdb",
+    "method": "git",
+    "tags": [
+      "library",
+      "bTree",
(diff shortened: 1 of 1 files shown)


dom96:
Your git repository has no ``.nimble`` file.

rivasiv:
Correct, but there is *.nims.   Is it obsolete? As I understood it may be used instead of *.nimble file. "A project.nims file can also be used as an alternative to a project.nimble file to specify the meta information (for example, author, description) and dependencies of a Nimble package. This means you can easily have platform specific dependencies." - I taked it from here : http://nim-lang.org/0.11.3/nims.html (Nimble integration section).

dom96:
That document is incorrect. The file extension has always been ``.nimble`` for Nimble. A ``.nims`` file can be executed by Nim and used to configure the build process, but that's it. The format of these files is the same though.

dom96:
I will merge it for now, but please do create a ``.nimble`` file and make sure that the package can be installed via ``nimble install``.

PR #429, opened 2016-11-27
Add gapbuffer

vktec:
Simple gap buffer implementation backed by `seq[char]`s. Not much more to it, really.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -5508,7 +5508,7 @@
       "seq",
       "sequence",
       "string",
-      "gap buffer"
+      "gapbuffer"
     ],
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -5498,5 +5498,20 @@
...
+  {
+    "name": "gapbuffer",
+    "url": "https://notabug.org/vktec/nim-gapbuffer.git",
+    "method": "git",
+    "tags": [
+      "buffer",
+      "seq",
+      "sequence",
+      "string",
+      "gap buffer"
```
dom96: No spaces in tags please.

PR #432, opened 2016-11-28
Update URL for `compiler` to point to nim-lang

vktec:
This isn't really an issue, since GitHub redirects, but it's confusing and a bit disconcerting to see nimble install from Araq/Nim.git instead of nim-lang/Nim.git


dom96:
Thanks :)

PR #439, opened 2016-12-16
Added package mutableseqs

iourinski:
Some methods imitating mutable sequences in scala etc.


dom96:
If you get a chance then please add some tags to your package, it will make it easier to find.

PR #444, opened 2016-12-29
simple stemmer

iourinski:
This is a simple stemmer for Russian and English language, there is a possibility to add new languages by providing a set of rules. It is a simple implementation of the algorithms described at http://snowballstem.org


dom96:
Nice! Thanks.

PR #452, opened 2017-01-12
Ajax package

define-private-public:
I'm probably going to start work on this next week, but I'd like to reserve the name right now.  It's going to be another light wrapper, this time over the `XMLHttpRequest` object.


yglukhov:
@define-private-public, you might want to have a look at https://github.com/yglukhov/jsbind and https://github.com/yglukhov/async_http_request

PR #454, opened 2017-01-16
Update packages.json

khchen:
Add winim package


dom96:
Nice, but why create this when the [windows](http://github.com/nim-lang/windows) package is available?

dom96:
> There is a oldwinapi for Windows API wrapper. But it is old and incompleted. So I start this project.

oh, I just saw your forum thread. Fair enough.

PR #457, opened 2017-01-18
Add libevdev, low-level Linux input processing

luked99:
This is just a small wrapper for libevdev, the Linux input processing library. I used it for handling the touchscreen on a Raspberry Pi.

It's possible it's not very "Nim-like", in that I just ran c2nim and then fixed up the errors. That means that the structures returned are all raw pointers, which should be fine if cleaned up using the API, but also means they can't be GC'd.


dom96:
Nice job :)

PR #462, opened 2017-02-01
Add gnuplot package

konqoro:


dom96:
Nice!

dom96:
btw please consider following the [Nim code style guide](http://nim-lang.org/docs/nep1.html).

Some important stylistic guidelines:

* https://github.com/dvolk/gnuplot.nim/blob/master/gnuplot.nim#L7 - Place the ``type`` on its own line.
* Only put a space after a ``:``.

dom96:
Oh, I see you've already made a PR for that. Nice :)

PR #466, opened 2017-02-09
Add package NimCx

qqtop:


packages.json
```diff
@@ -5892,5 +5892,16 @@
     "description": "Library for decompressing tar.gz files.",
     "license": "MIT",
     "web": "https://github.com/dom96/untar"
+  },
+  {
+    "name": "NimCx",
+    "url": "https://github.com/qqtop/NimCx",
+    "method": "git",
+    "tags": [
+      "yes"
```
dom96: can we get some better tags?

qqtop: I have been using nimble publish (for the first time) , so anything you see in package.json was not directly done by me, but you are right this needs improvement.
I  updated the package.json and send a new pull request. Hope this is ok now, if not my apologies.

dom96: Still looks the same to me.

PR #479, opened 2017-03-01
add snail and websockets

stisa:
- add [snail](https://github.com/stisa/snail) ( linear algebra in nim and js )
- add [websockets](https://github.com/stisa/websockets) ( websockets for nim js )

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -6003,5 +6003,32 @@
     "description": "Neverwinter Nights 1 data accessor library",
     "license": "MIT",
     "web": "https://github.com/niv/neverwinter.nim"
+  },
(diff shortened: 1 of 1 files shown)


rgv151:
there is already a `websocket` package, I think this package will makes ppl confusing.. it should be `websocketjs`.. instead

dom96:
I agree. Think you could rename it @stisa?

stisa:
Yes sorry, forgot to look for name clashes. I'll rename it to `jswebsockets` this afternoon.  
By the way, is there a convention for naming wrappers/bindings?

PR #490, opened 2017-03-31
Added cello

andreaferretti:
Added cello

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -6169,12 +6169,12 @@
     "method": "git",
     "tags": [
       "string",
-      "succint data structure",
+      "succinct-data-structure",
       "rank",
       "select",
       "Burrows-Wheeler",
-      "FM index",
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -6162,5 +6162,22 @@
...
     "license": "MIT",
     "web": "https://github.com/molnarmark/colorize"
+  },
+  {
+    "name": "cello",
+    "url": "https://github.com/unicredit/cello",
+    "method": "git",
+    "tags": [
+      "string",
+      "succint data structure",
```
dom96: Single word tags please

andreaferretti: I have used dashes, hope it is good. Many words do not really stand by themselves. For instance, this library provides wavelet trees, but it has nothing to do with wavelets (it is a historical accident that the former are named after the latter)

PR #500, opened 2017-05-05
add polyBool package

jangko:


dom96:
PolyBOOL is fine, but nimAGG isn't. Why does the git repo contain all of its dependencies?

jangko:
it's still in development process, and I am just lazy to remove the dependencies. they will be removed soon once it get more stable

dom96:
Okay, please make sure to do it in the future :)

PR #502, opened 2017-05-12
Add package sitmo

jxy:


packages.json
```diff
@@ -6288,5 +6288,19 @@
...
     "license": "MIT",
     "web": "https://github.com/jangko/freetype"
+  },
+  {
+    "name": "sitmo",
+    "url": "https://github.com/jxy/sitmo",
+    "method": "git",
+    "tags": [
+      "RNG",
+      "Sitmo",
```
dom96: For the future: this tag is redundant.

dom96:
thanks!

packages.json
jxy: I see.  Thanks.

dom96: No problem, Nimble should enforce that really :)

PR #504, opened 2017-05-18
Update package path for superfunc/{monad/maybe}

superfunc:
Hi Dom/All! Hope all is well!

I did some cleanup on my old package, including a name change, so here is  PR updating the path in nimble.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -1592,18 +1592,18 @@
     "web": "https://github.com/cowboy-coders/nim-openssl-evp"
   },
   {
-    "name": "monad",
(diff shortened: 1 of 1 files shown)


dom96:
From now on here is the rename procedure: https://github.com/nim-lang/packages#renaming-packages

superfunc:
😭 Sorry about that, didn't notice. Will keep it in mind in the future.

dom96:
Oh. No. It's a new thing, don't worry about it :)

PR #510, opened 2017-05-25
Add package getch

6A:


6A:
Heck, I only just realized there *is* a [`getch`](https://github.com/nim-lang/Nim/tree/master/lib/pure/terminal.nim#L627) method in Nim!

PR #515, opened 2017-06-02
Added serializetools.

grokqcd:
I've added a new package called serializetools. Thanks in advance for the help.


dom96:
Thanks :)

PR #522, opened 2017-06-15
nest take over

yglukhov:
Taking over [nest](https://github.com/kedean/nest). @kedean, please revert this when you feel like maintaining your lib.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -4351,7 +4351,7 @@
   },
   {
     "name": "nest",
-    "url": "https://github.com/kedean/nest.git",
+    "url": "https://github.com/yglukhov/nest.git",
     "method": "git",
     "tags": [
       "library",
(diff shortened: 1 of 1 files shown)


neeasade:
You are taking it because you the author didn't respond within less than 24 hours? What if they had other stuff going on/hadn't had time to review your PR yet? This feels premature.

yglukhov:
@neeasade, true, normally I would wait a bit longer, but (a) given that the lib doesn't compile because of changes in Nim that happened a couple of releases ago I would assume that @kedean is no longer interested in support, (b) judging from his github activity I would assume the same again, and (c) this particular PR can always be reverted with a single click and honestly I would be happy in case it is reverted.

yglukhov:
Resolved.

PR #534, opened 2017-07-06
Added NiGui

trustable-code:


dom96:
This file: https://github.com/trustable-code/NiGui/blob/master/src/msgbox.nim

Needs to be moved.

PR #538, opened 2017-07-13
Add package gintro

StefanSalewski:


packages.json
```diff
@@ -6723,5 +6723,16 @@
     "description": "Pack rects into bigger rect",
     "license": "MIT",
     "web": "https://github.com/yglukhov/rect_packer"
+  },
+  {
+    "name": "gintro",
+    "url": "https://github.com/stefansalewski/gintro",
+    "method": "git",
+    "tags": [
+      "yes"
```
dom96: Read before you answer "yes".

You need proper tags for your package.

dom96:
Nice to see that you're specifying the OS dependencies in your nimble file :)

If you want I can fix the tags for you.

StefanSalewski:
Yes, please fix the tags. Maybe something like GUI, GTK. You will know better.

The question for my answer yes was really not too well formulated, something like "Comma separated tag list?" I read carefully and was not sure if a yes/no answer was desired, or a actual tag list. Here in Germany, we would not use a question mark, when entering a list is desired.

PR #540, opened 2017-07-16
add sha3 library

tankfeeder:


dom96:
You should probably prevent the installation of those .rsp file by using the ``skipExt`` option (see nimble readme).

PR #548, opened 2017-07-26
Add package libgit2

barcharcraz:


barcharcraz:
I am not a smart man. Sorry about that. Nimble should probably refuse to publish in this case.

dom96:
Yeah, Nimble should. Anyway, let me know when you add a .nimble file and I will merge this.

PR #549, opened 2017-07-27
Add package oldgtk3

StefanSalewski:


dom96:
Why does your repo have so much other stuff that is skipped?

StefanSalewski:
>Why does your repo have so much other stuff that is skipped?

That are the scripts which I use for generating each Nim Module from the header files. I do provide them so everyone can generate the Nim modules new from scratch when desired. Useful for GTK versions > current 3.22 or when someone detects errors which can be fixed easier by modifying the generator script than by manually editing the Nim files. 

Exception is directory oldgtk3.bak, that one contains all the Nim modules in their original form with module names gtk3 and gdk3 -- for the final modules I applied a sed command to remove the "3" digit. That sed command is not reversible, so it is useful to keep the originals.

dom96:
You should just delete oldgtk3.bak, it will always be in the git history anyway.

PR #554, opened 2017-08-05
added calendar a clone of pythons calendar.py

skilchen:


dom96:
Nice! Any chance I can help with the commits? (you seem to have needed to make a lot).

btw, instead of putting examples into a ``calendarpkg`` directory you should just have it in the root directory and add ``skipDirs = @["examples"]`` to your .nimble file.

PR #555, opened 2017-08-07
Add package redux

pnuzhdin:


dom96:
Include more tags that are not the name of the package next time as well, for example ``library``, ``js`` :)

pnuzhdin:
@dom96 ok, but it's not limited by JS only 😄

PR #558, opened 2017-08-08
Highjack nim-gmp

subsetpark:
`nim-gmp` is an unmaintained package that's fallen out of compatibility with nimble. I've forked it and made a trivial change to restore compatibility.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -1665,7 +1665,7 @@
   },
   {
     "name": "gmp",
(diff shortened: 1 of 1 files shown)


dom96:
Beautiful, thank you!

@FedeOmoto let me know if you want to revert this :)

FedeOmoto:
That's OK!
Unfortunately, I don't have time to maintain all my Nim packages, so thanks for keeping it alive :smiley:

PR #563, opened 2017-08-10
nimrpc fresh repo

milisarge:


packages.json
```diff
@@ -3373,6 +3373,20 @@
     "license": "MIT",
     "web": "https://github.com/rogercloud/nim-rpc"
   },
+    {
+    "name": "nimrpc-milis",
```
dom96: Sorry but no dashes in package names.

dom96:
Great! You will also need to update your [repo](https://github.com/milisarge/nimrpc_milis/blob/master/nimrpc.nimble) (rename nimrpc.nimble to nimrpc_milis.nimble etc.). I'll merge this now anyway.

PR #566, opened 2017-08-15
Added package libgit2

barcharcraz:
I believe I've actually added the .nimble file. the issue was me being an idiot and forgetting to actually push the refs to my github repo.


dom96:
Thanks, looks good.

PR #573, opened 2017-08-26
Added package pylib

Yardanico:


Yardanico:
W: Another error attempting to request: https://bitbucket.org/mihailp/blake2/
  Error was: Name or service not known 
not my fault

PR #580, opened 2017-09-14
Added package csvql

Bennyelg:


packages.json
```diff
@@ -7170,5 +7170,23 @@
...
+    "url": "https://github.com/Bennyelg/csvql",
+    "method": "git",
+    "tags": [
+      "csv",
+      "read",
+      "ansisql",
+      "query",
+      "database",
+      "files",
+      "nim",
```
dom96: Please don't add these tags next time... every package is for Nim...

PR #583, opened 2017-09-15
Update packages.json

Bennyelg:
change package name,

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -7170,5 +7170,9 @@
     "description": "prestodb simple connector",
     "license": "MIT",
     "web": "https://github.com/Bennyelg/nimPresto"
-  }
+  },
+  {
+    "name": "db_presto",
+    "alias": "presto"
+  },
 ]


dom96:
You should keep the old name IMO.

Bennyelg:
I saw an example of how to add an alias in-order to let users to download the packages using the alias name and just did the follow,

here is an example from inside the package.json
```
  {
    "name": "monad",
    "alias": "maybe"
  },
  {
    "name": "maybe",
    "url": "https://github.com/superfunc/maybe",
    "method": "git",
    "tags": [
      "library",
      "functional",
      "optional",
      "monad"
    ],
    "description": "basic monadic maybe type for Nim",
    "license": "BSD3",
    "web": "https://github.com/superfunc/maybe"
  },
```

dom96:
Why do you want to rename it? I think the current name is fine.

Bennyelg:
@dom96 Because I want everyone install it `nimble install db_presto`
and use it `import db_presto`

dom96:
Okay, but why do you want this name and not "presto", what's wrong with ``import presto``?

Bennyelg:
because I wanted to save the conversion of importing databases in nim.
db_mysql
db_postgres
db_...

dom96:
Alright, for this case you can just rename the package directly (there is no need to alias since your package is very new). I'll do it for you.

PR #584, opened 2017-09-19
addded package contracts

Udiknedormin:
Added my Design by Contract package (under development but seems to work fine).

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -7200,5 +7200,28 @@
     "description": "csvql.",
     "license": "MIT",
     "web": "https://github.com/Bennyelg/csvql"
+  },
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -7200,5 +7200,28 @@
...
+  },
+  {
+    "name": "contracts",
+    "url": "https://github.com/Udiknedormin/NimContracts",
+    "method": "git",
+    "tags": [
+      "library",
+      "pure",
+      "contracts",
+      "design by contract",
```

Udiknedormin: Doesn't patty package have spaces in its tags?

PR #588, opened 2017-09-24
Add package huenim

truedat101:


dom96:
You've input the tags a little incorrectly but I can fix it :)

dom96:
Btw looks like an awesome package, thanks!

truedat101:
Thank you!  I do have a question about UDP multicast support.  a) Has anyone implemented UPnP or SSDP b) I wrote some code as part of this release that I couldn't get working as a an attempt to get a hue discovery going via UPnP using UDP.  I wasn't able to get this code working (no response from Hue) though I've implemented this same  protocol successfully on Java, C and node.js.  Should I ping you on the message boards about this question?

dom96:
You can ask here. If you ask on the forum or IRC though you might get answers from more people though :)

>Has anyone implemented UPnP or SSDP

Not that I'm aware of. 

As far as UDP support is concerned it needs improvement. Especially when it comes to async.

> I wrote some code as part of this release that I couldn't get working as a an attempt to get a hue discovery going via UPnP using UDP

Presumably you've used some low-level code for this. If it works in C then it should work in Nim. You might need to investigate the code in more detail to see why it behaves differently to the C code. I can take a look if you'd like.

truedat101:
@dom96 thank you.  I have created an issue here: https://github.com/IoTone/huenim/issues/3  .  This isn't a big show stopper since the Hue has a cloud service that uses something called nupnp.  It works, and solves the problem, however, it requires Internet access, which is a problem in demo situations where Internet may not be available, and therefore UPnP is a bettter solution.

PR #600, opened 2017-10-23
steal glm from stavenco

krux02:
Well this is basically stealing. But some time ago I actually discussed this with stavenco that he is not really active in Nim anymore and he would be ok if I would take this project over. If you look at the contributors list, you can see, that passing the ownership to me is justified, https://github.com/stavenko/nim-glm/graphs/contributors and it would simplify the process for me to improve this library.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -553,7 +553,7 @@
   },
   {
     "name": "glm",
-    "url": "https://github.com/stavenko/nim-glm",
+    "url": "https://github.com/krux02/nim-glm",
     "method": "git",
     "tags": [
       "opengl",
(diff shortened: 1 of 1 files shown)


stavenko:
Yep, since there are some pull request, I have no idea about, I cannot really move on on further library development.

dom96:
Weird, I thought I asked about this here (maybe on IRC). But again:

@krux02 maybe it would make more sense to just get @stavenko to add you as a contributor to their repo instead of forking because you'll lose all the issues/PRs

krux02:
ok, I became a contributor of glm, so this is solved. Thank you.

PR #611, opened 2017-11-06
Add zero_functional entry

alehander42:
Adding [zero-functional lib](https://github.com/alehander42/zero-functional) (is the name with `-` ok? I am not sure if there is a convention here). I can also rename it e.g. to `chain` or `chain-functional` if the name looks confusing


dom96:
You cannot have dashes in names, sorry.

dom96:
Your name needs to match your .nimble file.

alehander42:
Ok, so `zero_functional` is fine?

dom96:
Yep

PR #623, opened 2017-11-27
Added package normalize

nitely:


dom96:
Nice. Convention for Nimble is to use ``nimble test`` for tests btw.

PR #634, opened 2017-12-31
added avbin entry

Vladar4:


dom96:
You might want to simply call this package ``avbin``.

Vladar4:
Isn't it kinda misleading considering the existence of the original library repo? https://github.com/AVbin/AVbin

dom96:
I don't think so.

Vladar4:
Updated

dom96:
thank you

PR #636, opened 2018-01-05
add hts-nim

brentp:


packages.json
```diff
@@ -7319,6 +7319,22 @@
     "license": "Apache License 2.0",
     "web": "https://github.com/edubart/arraymancer-vision"
   },
+  {
+    "name": "hts-nim",
```
dom96: This name should match your .nimble file.

brentp:
changed and force-pushed. thanks.

PR #638, opened 2018-01-10
Added ndf

rustomax:


dom96:
Nice, beautifully written code too :)

PR #645, opened 2018-01-20
Add package shiori_charset_convert

Narazaka:


dom96:
That's a long package name, looks good though :)

PR #648, opened 2018-01-30
added grafanim

jamesalbert:


jamesalbert:
looks like a problem with jester

PR #660, opened 2018-02-24
Add useragents library.

treeform:


packages.json
```diff
@@ -8256,5 +8256,17 @@
...
     "license": "MIT",
     "web": "https://github.com/dom96/pwned"
+  },
+  {
+    "name": "useragents",
+    "url": "https://github.com/treeform/useragents",
+    "method": "git",
+    "tags": [
+      "library",
+      "user agent"
```

PR #662, opened 2018-02-26
Add package suffer

emekoi:


dom96:
You should put this directory in a ``suffer`` directory: https://github.com/emekoi/suffer/tree/master/src/private

It does only contain .c files, but it's good to be consistent with Nim I think.

dom96:
oh, I didn't realise you skipped it. That's fine. :)

PR #663, opened 2018-02-27
Register ttmath

alehander42:
Yea


dom96:
Sorry to be picky, but your package pollutes the namespace with this file: https://github.com/status-im/nim-ttmath/blob/master/src/ttmathuint.nim

Use ``nimble check`` to verify your package's structure.

PR #676, opened 2018-03-11
Add package moduleinit

skunkiferous:


packages.json
```diff
@@ -8593,5 +8593,16 @@
     "description": "Text editor to speed up testing code snippets",
     "license": "MIT",
     "web": "https://github.com/genotrance/snip"
+  },
+  {
+    "name": "moduleinit",
+    "url": "https://github.com/skunkiferous/moduleinit",
+    "method": "git",
+    "tags": [
+      "v0.1.0"
```
dom96: This isn't where you put git tags, Nimble can read those independently from your Git repo.

The tags are just categories, see other package's tags.

skunkiferous:
Hi,

could you maybe send me the build error? It builds on my machine, ans I
don't have a Mac to test on...

C:\Code\moduleinit>nimble build
  Verifying dependencies for moduleinit@0.1.0
   Building moduleinit/moduleinit.exe using c backend

C:\Code\moduleinit>

Also, if I'm not meant to put git tags in tags, what am I meant to put in
tags? Leaving it empty is not an option.

skunkiferous:
OK, I guess "library" would be the tag to use.

Could the line "--threads:on" im my default Nim installation be causing the
problems? (I have to guess, since I don't have the error...) I'll add it to
the package's nim.cfg and try again.

On Sun, Mar 11, 2018 at 2:10 PM, Sebastien Diot <skunkiferous@googlemail.com

dom96:
Looks like it's building successfully now

skunkiferous:
Unrelated to my new package, I wanted to see a "human readable" version of
all nimble packages (to make sure I don't choose a name already taken for
my next package), and I think there's a problem there. On this page:

https://nim-lang.org/docs/lib.html

I think they should be displayed at the bottom, but all I see is:

"If you are reading this you are missing nimblepkglist.js or have
javascript disabled in your browser."

I tried in 2 different browsers. I checked if JS is activated in my browser
there:

Is JavaScript Enabled?
<https://www.whatismybrowser.com/detect/is-javascript-enabled>

and the site says "yes", so I think something is broken.

dom96:
Yes, we need to fix it: https://github.com/nim-lang/website/issues/74

dom96:
Superseded by another PR.

PR #677, opened 2018-03-11
Add package moduleinit

skunkiferous:


dom96:
Thanks for fixing! Added some more tags as well.

PR #680, opened 2018-03-14
Added package nimbigwig

genotrance:


dom96:
Nice, btw pretty sure you don't need the distros module to check for Windows. You can just do ``when defined(windows)``.

PR #681, opened 2018-03-15
Added package regex

nitely:


dom96:
Nice. Next time try to pick tags which are not just the same as the name of the package though :)

PR #685, opened 2018-03-18
Add package nimclutter

KeepCoolWithCoolidge:


dom96:
I wonder why the GTK3 package is called ``oldgtk3``, is there a new GTK3 package?

StefanSalewski:
I think we all know the story, at least we have discussed it multiple times.

First there was ngtk3 with single modules without nimble support.

People asked for nimble support, but most package names where already blocked with araqs legacy modules, for example cairo, pango, glib...

Then I started an gobject-introspection high level package, named nim-gi. But that name was not really loved, as nim prefix and hypen was deprecated. Araq suggests name gintro -- this is still the most interesting package.

But as some people may still want a low level package I have updated ngtk to latest GTK3 version and bundled it into oldgtk3. oldgtk3 is a unique name, and at the same times indicates that there is something more advanced, which is gintro.

dom96:
``oldgtk3`` indicates that there is a newer gtk3 wrapper. IMO you should call it ``gtk3wrapper``, or just ... ``gtk3`` (it's available AFAIK)?

StefanSalewski:
gtk3 would not be a good top level package term at all, because the package includes, and has to include cairo, pango, glib2 and similar, which are not closely related to gtk3. We have discussed all that in  the last three years, so that ship has sailed. And still we have only one serious GTK3 user -- that gui is smart, so he is not confused by names.

Maybe for upcomming GTK4 we can find better names -- I still have no idea.

dom96:
I don't think we discussed it enough then. But this really shouldn't be hard. 

Looking at your ``oldgtk3`` I am really puzzled and concerned. It contains a number of directories which you are skipping in the Nimble file, why are they there then? Why do you have a .bak folder committed into Git?

Your argument for not naming it ``gtk3`` can also be applied to ``oldgtk3``... 

And looking at your ``gintro`` package, you're copying files from ``oldgtk3`` [during installation](https://github.com/StefanSalewski/gintro/blob/master/gintro.nimble#L43)?! What!?! Why are you doing this?

StefanSalewski:
Yes, I think with the .bak directory you are right, I can delete it, I think you told me some months ago. I made a bak to still have access to older release, but I think git will gave me that access too, and no one complained about latests commits, so I will indeed delete the bak in the next days.

For for your last question, we really discussed that on IRC enough.

dom96:
> For for your last question, we really discussed that on IRC enough.

Do you have a link to when we discussed this? Could we revisit it? I'm sorry, but I don't remember discussing this with you.

StefanSalewski:
A recent discussion was this:
```
https://irclogs.nim-lang.org/18-11-2017.html
20:48:27 | * | salewski joined #nim

salewski:Can we use wget in nimble files? Or may that be not available for mac and windows?
salewski: Because I would like to use oldgtk3 to bootstrap ginto, but I want not to clone full oldgtk3 wich user does not need later.
```

But that one was not the last, and not the first.

The core of the problem is, that gintro generates all the files from a XML like database using gobject-introspection. Generated files are OS dependant, so it makes not much sense to ship pregenerated files. For generation, 3 modules are necessary, including gobject.nim, glib.nim and gir.nim. Initially I took them from ngtk3 repository, but as that one is going to be deleted because it is obsolete and deprecated, I now take these 3 files from oldgtk3 package. Using wget or nimgrab. Instead of this, I could force a full oldgtk3 install as a prerequisite for gintro install. Or I could just copy the 3 files into gintro package. Copy is simple, but some sort of lame.

dom96:
So it looks like you discussed it with Araq, not me. 

> Instead of this, I could force a full oldgtk3 install as a prerequisite for gintro install.

This is what you should do. What you're doing is a hack and for what, a few MB of space savings (if even that)? If you really want leaner packages then split up your ``oldgtk3`` package into ``glib3``, ``gtk3``, etc. You can even have all of these in the same repository now (Nimble now supports [this](https://github.com/nim-lang/nimble/#package-urls) since its latest release)

StefanSalewski:
>your oldgtk3 package into glib3, gtk3, etc.

We have discussed these names -- a glib3 does not exist from upstream, same for pango, cairo and others. So glib-salewski, cairo-salewski, pango-salewski was an ugly option. All that is very confusing for users. When nimble offers now new possibilities, then I will investigate these. The now working "before-install" hook was indeed a big step forward for gintro package, as with it an easy "nimble install gintro" was possible for the first time. And that seems to work for Windows and Linux!

I can not say when I will have time to work on ginto or oldgtk3 again, my motivation will of course depend on number of users, currently there seems to be between 0 and 2, which is not that much.

PR #689, opened 2018-03-25
Added package sol

davidgarland:
Attempting to add my vector library Sol to Nimble.

If I did anything incorrectly, let me know. :)


davidgarland:
Actually-- I realized the project has a huge flaw with how it does SIMD without vectorcall convention and such.

I'll wait on this.

PR #693, opened 2018-03-29
Add package sol

davidgarland:
OK, fixed the problem I had last time. Everything uses GNU vector intrinsics internally now, so there aren't any major performance deficiencies compared to scalar-- it should be much faster than scalar code, as SIMD should be.

Again, if I did anything incorrectly, such as package structure, let me know.


dom96:
Looks good!

PR #694, opened 2018-03-30
Add protobuf and combparser packages

PMunch:


dom96:
No dashes allowed in package names. Also please add new packages at the bottom.

PMunch:
Oh woops, didn't mean to put the dash in there, was just doing a multi-replace for github URL, website, and name.

dom96:
Some remarks:

For combparser you should really use the ``options`` module instead of creating your own ``Maybe`` type. Although I see your ``Maybe`` type is in fact an ``Either`` type. Consider renaming :)

I wonder what we should do about @oswjk's package? https://github.com/oswjk/protobuf-nim. It seems you submitted this PR first so I guess you win this name, but it still sucks a bit for @oswjk, any thoughts?

PMunch:
Yeah, combparser should really be rewritten from scratch. It was all based [this](https://gist.github.com/kmizu/2b10c2bf0ab3eafecc1a825b892482f3) two year old snippet written in a very functional style, and not very Nimish (Nim-esque? Nim-like? Nim-tastic? Nim-tacular? What do we use to describe things that are "Nim", similar to "Pythonic"). The original Maybe type should be renamed something else entirely, and the entire structure should be given a second (or third) thought. I just needed to publish it to get protobuf published.

I was wondering the same thing about @oswjks package, actually a bit relieved to find he didn't beat me to it so I didn't have to think up a better name :P I'd be happy to have none of them use "protobuf" though, and let them both have some descriptive suffix. But it's hard coming up with some short suffix that accurately describes the difference between them. A second option is of course to merge them into a super-package. If the protoc plugin could spit out an object initialisation for my ProtoNode object structure you would be able to write a similar function to what he's doing and generate the same code as I'm doing now.

oswjk:
I'm okay with you taking the name :-)

PMunch:
Oh sorry, I fixed it ages ago but just forgot to push :P I was wondering why it took you so long to merge

PR #695, opened 2018-04-01
added simdX86

jcosborn:


Yardanico:
@jcosborn you can also check https://github.com/jackmott/nim_simd because it implements simd instructions too

dom96:
Perfectly structured package. Thanks!

jcosborn:
@Yardanico the scope of that is higher level.  He uses the wrappers from https://github.com/bsegovia/x86_simd.nim, which are nice, but seem to be hand written and I wanted to add AVX512 and overloading, so starting with the C prototypes and using c2nim seemed to be the best way to go.  Hopefully this is suitable as a replacement for any other low-level wrappers people are using.

PR #697, opened 2018-04-04
Add package tinamou

Double-oxygeN:


dom96:
Installation fails for me with:

```
 ... tinamoupkg/core.nim(60, 39) Error: undeclared identifier: 'MIX_DEFAULT_FORMAT'
```

There is no reason to provide this binary. You should just remove it from ``bin`` and not use the ``pkg`` suffix for your package's directory.

Double-oxygeN:
Hmm... This error seems to be strange for me because `core.nim` was compiled successfully in my environment.  
I eliminated the `pkg` suffix and slightly changed `core.nim`. Would you check it again?

dom96:
Ahh, the issue was I had an outdated ``sdl2`` installed. Nimble needs lock files :)

You still shouldn't have this binary though, is it just to test compilation of your package? Please don't do it this way.

Double-oxygeN:
No, this project was created in order to develop games easily in Nim. The reason why I would like to publish it is that it will become easy to install it by using nimble, especially for my development team. And it is multi-genre, so it could be also useful for others.
I agree that this library is incomplete yet. If you say it too and the package is worthless for nimble, please reject my proposal. Then I will try to make it more mature.

dom96:
Yeah, your library is great. The Nimble file just shouldn't contain this ``bin`` option.

Double-oxygeN:
Oh, I understand. (My apologies for the lack of my English skills...)
What I have to do is to delete the lines about binary files from `tinamou.nimble`, isn't it?

dom96:
yep

Double-oxygeN:
I fixed the problem.

dom96:
Brilliant, thanks!

PR #698, opened 2018-04-04
Add package consul

subsetpark:


dom96:
Nice, but please please pick better tags next time. Something that gives people an idea what consul is.

PR #709, opened 2018-04-24
Add nimpb and nimpb_protoc

oswjk:
I'm adding these two packages in the same PR because nimpb kind of needs nimpb_protoc to be useful.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -8988,5 +8988,32 @@
     "description": "Test failing snippets from Nim's issues",
     "license": "MIT",
     "web": "https://github.com/genotrance/tissue"
+  },
+  {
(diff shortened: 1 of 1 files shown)


dom96:
Looks good, nice :)

dom96:
In case it helps, Nimble supports multiple packages in the same repo now :)

https://github.com/nim-lang/nimble#package-urls

oswjk:
Thanks for mergin :-)

I think I'll keep those packages separate. If a user already has protoc, then they wouldn't need the support package. And it's nicer to keep the protoc binaries in another repo, so if you don't need them, you'll save some download time.

PR #711, opened 2018-04-26
Added turn_based_game to list of packages.

JohnAD:
The library `turn_based_game` is a rules engine that assists in writing games that are...well...turn-based. (The meaning of that is well documented in the wiki.) One of my next projects is porting my Negamax AI algorithms to nim, and they will require this framework to predictably operate.


dom96:
Nice! You can use Nim syntax highlighting in your readme btw :)

PR #712, opened 2018-05-02
Added package translation

juancarlospaco:


dom96:
So which is it? ``tinyslation`` or ``translation``?

juancarlospaco:
The exported proc is `tinyslation`, the package `translation`. :smile_cat:

dom96:
A bit inconsistent, but oh well.

juancarlospaco:
Ok, next time I will use `tinyslation.tinyslation` like. :smiley_cat:

dom96:
IMO ``tinyslation.translate`` would make most sense :)

PR #713, opened 2018-05-04
Added package random_font_color

juancarlospaco:
https://github.com/juancarlospaco/nim-random-font-color#nim-random-font-color


dom96:
Only slightly better than ``leftpad`` but alright :)

PR #723, opened 2018-05-05
Add package nimhttpd

h3rald:


dom96:
screen shot 2018-05-05 at 18 33 47

h3rald:
Yes but... I specified that file within **skipFiles** as requested, what would be the best place for a nakefile otherwise?

dom96:
It seems like you can put all of that into your .nimble file (you're still using the old .nimble file format too). In either case, the nakefile shouldn't be installed.

h3rald:
I see! I missed the new format somehow... However, if I do that I cannot import the nimble file in the main file because the modules have the same name (nimhttpd.nim and nimhttpd.nimble), and I need a way to store the program name and above all the version number in a single file.

Would the whole nimble packaging still work if I put the package details (name, version, description, etc.) in another file and then import it from nimhttpd.nimble?

h3rald:
There... version 1.0.5 should be properly "nimble-compliant". Using NimScript for the nimble file made things simpler, BUT now I have to update the version number in two different files (nimhttpd.nimble and nimhttpd.nim). Any idea on how to solve this?

dom96:
> Would the whole nimble packaging still work if I put the package details (name, version, description, etc.) in another file and then import it from nimhttpd.nimble?

Yep

dom96:
Hrm, now your package isn't building for me.

h3rald:
I see... OK, I modified the project structure and tested it with nimble check and nimble build. Everything should be in order now...

dom96:
Thanks :)

PR #724, opened 2018-05-06
Add package crc32

juancarlospaco:
https://github.com/juancarlospaco/nim-crc32#nim-crc32


dom96:
If you copied it from Rosetta Code then it would be nice to use a more permissive license like MIT ;)

juancarlospaco:
Changed to MIT.

dom96:
thx

PR #729, opened 2018-05-08
Renamed mpint to stint

yglukhov:


dom96:
https://github.com/nim-lang/packages#renaming-packages :)

PR #733, opened 2018-05-11
Add package glfw

CodeDoes:


dom96:
This package's nimble file needs to be renamed: https://github.com/ephja/nim-glfw/blob/master/nim_glfw.nimble

Is there a reason @ephja isn't submitting it?

CodeVance:
Yes. @ephja hasn't been on github for a while. I've also tried making a PR
with his repo, but as of yet have not recieved any reply.

ephja:
Isn't the package already in there? Anyway, the format seems to be correct now

dom96:
Yes, this package is already there, but with a wrong name: ``nim-glfw``. Can either of you make a PR to rename it? 

https://github.com/nim-lang/packages#renaming-packages

ephja:
@CodeDoes I see that you previously tried to publish the package, but it was already present in the database, so that's why it didn't work. So all that remains is renaming the package; done in #773

CodeDoes:
Should I close this?

PR #736, opened 2018-05-12
jdec module

diegogub:
Hello,
I coded this module with helper functions for json marshal/unmarshal. To do things I could not with current json support.

Thanks!


dom96:
Why not improve the json module instead? :)

PR #740, opened 2018-05-15
Add package about

aleandros:


dom96:
Not sure if your intention is for this to be a library and a binary package. You can just write ``skipExt = @["nim"]`` in your .nimble file and then you won't need this weird ``aboutpkg`` folder (you'll be able to rename to ``about``)

PR #742, opened 2018-05-18
Added package findtests

jackvandrunen:


packages.json
```diff
@@ -9356,5 +9356,21 @@
...
+  },
+  {
+    "name": "findtests",
+    "url": "https://github.com/jackvandrunen/findtests",
+    "method": "git",
+    "tags": [
+      "test",
+      "tests",
+      "testing",
+      "unit test",
```
dom96: no spaces in tags please

dom96:
Otherwise looks good

jackvandrunen:
Saw that some other packages had it so thought it was ok. All fixed now!

dom96:
What other packages have it?

jackvandrunen:
From a quick glance it seems like quite a few of the earlier packages have them. I specifically saw them in einheit because I stole some unit-test-related flags, but you can see that there are at least a couple other packages around that area which have the same issue.

PR #745, opened 2018-05-20
Add package negamax

JohnAD:


dom96:
Nice! Looks good. In case you weren't ware, you can use Nim syntax highlighting in your readme with \```nim ... \```.

PR #755, opened 2018-05-29
Add sctp and sodium

zielmicha:


dom96:
```
    Warning: Package 'sctp' has an incorrect structure. It should contain a single directory hierarchy for source files, named 'sctp', but file 'test_send_congestion.nim' is in a directory named 'tests' instead. This will be an error in the future.
      Hint: If 'tests' contains source files for building 'sctp', rename it to 'sctp'. Otherwise, prevent its installation by adding `skipDirs = @["tests"]` to the .nimble file.
```

These need to be fixed.

dom96:
Also got this error when installing sctp:

```
Error: Downloaded package's version does not satisfy requested version range: wanted >= 0.4.6 got 0.4.5
```

zielmicha:
Oops, I forgot to make some pushes. It works now.

PR #761, opened 2018-06-10
Add package nimsvg

magick93:


dom96:
Cool thanks. Hope @bluenote10 isn't keeping this out of the packages list on purpose.

PR #785, opened 2018-07-06
Add package fmod

johnnovak:


dom96:
screen shot 2018-07-04 at 10 25 11

dom96:
Huh, turns out that's a bug in my installation.

PR #795, opened 2018-07-13
more animation packages

Skrylar:


dom96:
Nimble packages need a .nimble file...

PR #802, opened 2018-07-18
Add package timecop

ba0f3:


dom96:
Seems Travis had a nice fail with this PR.

screen shot 2018-07-18 at 20 59 32

PR #812, opened 2018-08-04
Added nim-tiled

SkyVault:
This is a simple yet really useful library I wrote to load tmx files generated by the very popular tiled tile map editor, https://www.mapeditor.org/


Yardanico:
You created two PRs btw - https://github.com/nim-lang/packages/pull/811/files

PR #844, opened 2018-08-27
Remove package monero

ehmry:
Repo has been removed from github.


ehmry:
It was incomplete and I do not plan to continue developing it. Someone has contacted me about it and I have given them the source, so they may choose to add it again later.

PR #845, opened 2018-08-27
Add package finalseg

bung87:


dom96:
Next time: specify the tags (see the tags I added for an example)

bung87:
never aware of it before you add it, seems there's no a page like this https://pypi.org/classifiers/
I'll find proper word taged it next time.

PR #851, opened 2018-09-04
Update packages.json

khchen:
Nim's Windows GUI Framework.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -10597,11 +10597,11 @@
       "library",
       "windows",
       "gui",
-      "ui"      
+      "ui"
     ],
     "description": "Nim's Windows GUI Framework.",
     "license": "MIT",
     "web": "https://github.com/khchen/wNim",
-    "doc": "https://khchen.github.io/wNim/wNim.html'
+    "doc": "https://khchen.github.io/wNim/wNim.html"
   }
 ]
--- a/packages.json
+++ b/packages.json
@@ -10588,5 +10588,20 @@
     ],
     "description": "Event Based Programming for Nim.",
     "license": "MIT"
+  },
+  {
+    "name": "wNim",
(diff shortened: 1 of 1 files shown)


dom96:
```
   Warning: Package 'wnim' has an incorrect structure. It should contain a single directory hierarchy for source files, named 'wnim', but file 'wColors.nim' is in a directory named 'wNim/private/consts' instead. This will be an error in the future.
      Hint: If 'wNim/private/consts' contains source files for building 'wnim', rename it to 'wnim'. Otherwise, prevent its installation by adding `skipDirs = @["wNim/private/consts"]` to the .nimble file.
```

khchen:
I can install locally without warning?

nimble install
Verifying dependencies for wNim@0.1.2
  Info: Dependency on winim@>= 2.4.2 already satisfied
Verifying dependencies for winim@2.4.3
  Info: Dependency on kiwi@>= 0.1.0 already satisfied
Warning: Package 'kiwi' has an incorrect structure. It should contain a single directory hierarchy for source files, named 'kiwi', but file 'common.nim' is in a directory named 'tests' instead. This will be an error in the future.
  Hint: If 'tests' contains source files for building 'kiwi', rename it to 'kiwi'. Otherwise, prevent its installation by adding `skipDirs = @["tests"]` to the .nimble file.
Verifying dependencies for kiwi@0.1.0
Installing wNim@0.1.2
Success: wNim installed successfully.

And this project already have single directory hierarchy for source files, named 'wnim'.
What is the problem?

dom96:
Huh, strange. It seems like there is some regression in latest Nimble that's causing this warning. I'll look into it.

khchen:
Maybe it is the problem about letter case? 
Path should be case insensitive in Windows but nimble don't recognize it.

PR #852, opened 2018-09-06
add genoiser

brentp:


dom96:
Note:

```
githubcom_brentphtsnim_0.2.0/src/hts/vcf.nim(71, 36) Error: type mismatch: got <nil> but expected 'seq[string]'
```

You'll need to fix that for upcoming 0.19.0.

brentp:
thanks for merging. I've been going through my various projects and updating to work on nim@devel including the vcf.nim file that you mention.

PR #861, opened 2018-09-18
Fix psutil

juancarlospaco:
Fix psutil, point to new maintained fork, original author credits remain, an explanation is also added, 
author is not even active on github at all, if original author ever comes back with a maintained repo feel free to revert back,
no drama lama, I just want to see things working. All Tests pass Ok for `stable` and `devel`.
:cat:

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -5849,7 +5849,7 @@
   },
   {
     "name": "psutil",
-    "url": "https://github.com/juancarlospaco/packages",
+    "url": "https://github.com/juancarlospaco/psutil-nim",
     "method": "git",
     "tags": [
       "psutil",
--- a/packages.json
+++ b/packages.json
@@ -5849,7 +5849,7 @@
   },
   {
     "name": "psutil",
-    "url": "https://github.com/johnscillieri/psutil-nim",
+    "url": "https://github.com/juancarlospaco/packages",
     "method": "git",
     "tags": [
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -5849,7 +5849,7 @@
   },
   {
     "name": "psutil",
-    "url": "https://github.com/johnscillieri/psutil-nim",
+    "url": "https://github.com/juancarlospaco/packages",
```
dom96: This URL doesn't look right

dom96:
Merging, but please fix the directory structure:

```
   Warning: Package 'psutil' has an incorrect structure. The top level of the package source directory should contain at most one module, named 'psutil.nim', but a file named 'common.nim' was found. This will be an error in the future.
      Hint: If this is the primary source file in the package, rename it to 'psutil.nim'. If it's a source file required by the main module, or if it is one of several modules exposed by 'psutil', then move it into a 'psutil/' subdirectory. If it's a test file or otherwise not required to build the the package 'psutil.nim', prevent its installation by adding `skipFiles = @["common.nim"]` to the .nimble file. See https://github.com/nim-lang/nimble#libraries for more info.
```

juancarlospaco:
Yes, Ive seen that too, the idea was to make a 1st one with smallest possible set of changes to make it work, then fix and add features.

PR #866, opened 2018-09-24
Add torch package variation to nimtorch

sinkingsugar:
We renamed the `.nimble` file into `torch.nimble`

Would be wise to keep both flavors the old `nimtorch` (not so correct API wise but works) and the new flavor `torch`


sinkingsugar:
Should be fixed! thanks for pointing this out! @dom96

PR #880, opened 2018-10-04
Repository relocation

lguzzon:


dom96:
okay, would be nice if you used a different avatar for your org.

lguzzon:
Sure, this could be fine?
Or it's better to no "refers" to NIM logo?
lguzzon-nim

Issue #399, opened 2016-08-17
Tag naming / checking


abudden:
Having just started adding my package to the package list, I was looking at the existing tags to see which were most appropriate.  I found a lot of duplication.  A few examples from the start of the alphabet:
- "2d", "2D"
- "authentication", "authenticator"
- "algorithm", "algorithms"
- "app", "application"
- "benchmark", "benchmarking"
- "bind", "binding", "bindings"
- "Bloom", "Bloom filter"

The list goes on (and even includes "Nim", "nim" and "nimrod"!)...

If these tags are ever selected as the means of trying to search/filter packages, it's going to be very hard to actually find anything useful I suspect.

Araq:
Stemming and case insensitive searches are not rocket science... I don't see any problems here.

dom96:
PRs to clean up tags are always welcome!

achesak:
At the very least, can "nim" and "nimrod" be removed as tags? They're rather pointless as every package is going to involve Nim anyway.

dom96:
Sure

PR #883, opened 2018-10-09
Add package witai

xmonader:


packages.json
```diff
@@ -10992,5 +10992,19 @@
...
     "license": "MIT",
     "web": "https://github.com/genotrance/nimtess2"
+  },
+  {
+    "name": "witai",
+    "url": "https://github.com/xmonader/witai-nim",
+    "method": "git",
+    "tags": [
+      "wit.ai",
+      "nim",
```
kaushalmodi: The "nim" tag is unnecessary. See https://github.com/nim-lang/packages/issues/399#issuecomment-243021190, https://github.com/nim-lang/packages/pull/886.

PR #895, opened 2018-10-15
unpublish nimcalcal, calendar

skilchen:
my packages are outdated and unmaintained.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -7345,18 +7345,6 @@
     "web": "https://github.com/trustable-code/NiGui"
   },
   {
-    "name": "nimcalcal",
-    "url": "https://github.com/skilchen/nimcalcal",
-    "method": "git",
-    "tags": [
-      "calendar",
-      "library"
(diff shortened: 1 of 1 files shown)


dom96:
I would prefer if you state in your repo that your package is unmantained. Let others use them if they wish.

Araq:
Well once published, others may depend on it. Nimble needs to fork all packages or we need some other mechanism.

skilchen:
The packages are no longer usable and I won't fix them. So let them disappear from the published package list.

dom96:
It's better to give people something to fix than to make the package disappear completely.

skilchen:
One advantage of unpublishing: the names of the retired packages can be used to publish better alternatives.

Your concerns about disappearing packages are valid. 

But for my packages: 
- i now think it was a mistake to publish them
- it really doesn't matter if they are no longer visible in nimbles package list

So please merge my PR.

PR #899, opened 2018-10-17
Add package markdown

soasme:


soasme:
@dom96 is there any package naming rules that I should follow?

dom96:
Looks beautiful ;)

Only thing I would suggest is to get rid of the makefile and use Nimble tasks instead.

PR #901, opened 2018-10-20
Add nimtomd

ThomasTJdev:
Converts Nim code to Markdown. Can convert a whole Nim file or a string with Nim code.

https://github.com/ThomasTJdev/nimtomd


ba0f3:
awesome, this is what I was looking for

PR #902, opened 2018-10-20
Add sugerror

Quelklef:
:kiss:

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -11156,7 +11156,7 @@
     "tags": [
       "error",
       "errors",
-      "error handling"
+      "expr"
     ],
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -11148,5 +11148,18 @@
...
     "web": "https://github.com/soasme/nim-markdown"
+  },
+  {
+    "name": "sugerror",
+    "url": "https://github.com/quelklef/nim-sugerror",
+    "method": "git",
+    "tags": [
+      "error",
+      "errors",
+      "error handling"
```

krux02: doesn't nimble have a builtin linter for tags?

PR #904, opened 2018-10-21
Add package Coco

samuelroy:
Coco is a code coverage CLI & library for Nim


samuelroy:
It's fixed now. Btw it's really weird that skipDirs wasn't working as expected. I had to use installFiles instead.

PR #916, opened 2018-10-31
Refactor package scanner

juancarlospaco:
# Refactor the package scanner

**Checks for:**

- if the Git Repos actually exist.
- if the `*.nimble` file actually exist.
- if the `Alias` relate to an existent `Name`.
- Blacklisted tags, like `nim`, `nimrod`, etc.
- Empty strings on fields.
- Unwanted whitespaces on tags.
- Too much tags, too few tags.
- Packages with duplicated URLs.
- Enforce required keys.
- Maximum number of tags.
- Maximum length of tags.
- More precise checking of the JSON.
- Other minor cases observed on the current JSON.

**Other features:**

- Updated for Nim `0.19.0`.
- Everything can be Configured or Disabled from a `const`.
- Self-documented, now `nim doc` works (to attract more contributions).
- Works online/offline, ( `when defined(ssl)` it will go Online and check all repos).
- Uses proper `logging` instead of `echo`.
- Overall is an improvement over the previous one.

**To be defined:**

So this "Linter" finds more errors on the current JSON, I see 2 options:

- Enable all checking, fix the JSON errors.
- Disable some checking, keep the JSON with some errors.

So basically you can pick and choose on the Strictness.

**Feel free to use the Suggestions feature of GitHub to suggest improvements.**
:cat:


dom96:
Things I see immediately: camelCase and also test shouldn't fail.

But this is great, thank you!

juancarlospaco:
Tests are expected to fail, because of the mentioned imperfections on the current JSON, as explained above.
:slightly_smiling_face:

genotrance:
What are the next steps to get this merged?

Having a report of all issues for every package will be helpful - then each package owner can be asked to update their package. If no update is provided in a stipulated amount of time, we can then disable it.

wizzardx:
> Feel free to use the Suggestions feature of GitHub to suggest improvements.

I searched for but couldn't find the Suggestions GitHub feature.

I'll add a suggestion or idea over here instead (for now). Some kind of ongoing CI against dev and stable (and maybe stable-1) for the packages.

This might make more sense after Nim 1.0 comes out.

Some kind of auto-maintained record of "last known to be good" combination of packages and compiler versions would also be a very nice artifact to generate - I'm thinking along the lines of how Haskell's Stackage service works, and their "nightly" snapshots).

There's probably some overlap between this idea, and what the nimble.directory service already does. eg, if we look at NimBorg on there:

https://nimble.directory/pkg/NimBorg

juancarlospaco:
@dom96 Can you check again please :question: 
is all updated for `0.20.0` and green. :slightly_smiling_face:

package_scanner.nim
```diff
@@ -1,135 +1,367 @@
...
+## - Alias relate to Names
+## - Profanity Filter by Levenshtein Distance.
+## - Try to download the nimble file from the repo.
+## - Try to Git clone the repo (deletes folder after).
+## - Works online/offline
+##
+## Use
+## ---
+##
+## - Off-line: ``nim c -r package_scanner2.nim`` (Faster, wont use Internet).
```
dom96: I'd rather this be an explicit option, silently disabling internet checks could cause problems down the road.

package_scanner.nim
```diff
@@ -1,135 +1,367 @@
...
+    "public domain",
+    "lgplv3 or gplv2",
+    "openssl and ssleay",
+    "openldap",
+    "apache 2.0 or gplv2",
+    "mit or apache 2.0",
+    "lgpl with static linking exception",
+  ]  ## All valid known licences for Nimble packages, on lowercase.
+when checkProfanity:
+  const profanitiesStr = readFile(currentSourcePath().parentDir / "profanities.txt").normalize.splitLines
```
dom96: This is cool, but I don't think it's necessary, in fact if people want to call their package "fuck" I have no problem with that. I think humans can reject anything more offensive.

package_scanner.nim
```diff
@@ -1,135 +1,367 @@
...
+addOutputFormatter(defaultConsoleFormatter())
+addOutputFormatter(report)
+when checkProfanity:
+  let profanities = filterIt(profanitiesStr, it.len > 0 and not(it.startsWith"#"))
+
+proc handler() {.noconv.} =
+  quit("CTRL+C Pressed, package_scanner is shutting down, Bye.")
+setControlCHook(handler)
+
+using url, name: string
```
dom96: I personally dislike this construct, we shouldn't be using it.

package_scanner.nim
```diff
@@ -1,135 +1,367 @@
...
+  rubyCmd = "ruby ./validate_json.rb"
+  defaultFilePermissions = {fpUserWrite, fpUserRead, fpGroupWrite, fpGroupRead, fpOthersRead}
+  packagesFilePath = currentSourcePath().parentDir / "packages.json"
+  packagesJsonStr = readFile(packagesFilePath)
+  hostsSkip = [
+    "https://bitbucket",
+    "https://gitlab.3dicc",
+    "https://mahlon@bitbucket",
+    "https://notabug",
+  ] ## Hostnames to skip checks, for reliability, wont like direct HTTP GET?.
```
dom96: These are strange URLs, they don't contain a real hostname. I assume you are parsing them in some custom way? Why not make this more clear with:

```nim
hostsSkip = [
  "bitbucket.org",
  "gitlab.3dicc",
  ...
]
```

package_scanner.nim
```diff
@@ -1,135 +1,367 @@
...
+    "apache license 2.0 or mit",
+    "fontconfig",
+    "gnu lesser general public license v2.1",
+    "public domain",
+    "lgplv3 or gplv2",
+    "openssl and ssleay",
+    "openldap",
+    "apache 2.0 or gplv2",
+    "mit or apache 2.0",
+    "lgpl with static linking exception",
```
dom96: We need to be more strict and possibly implement a license parser which understands things like "MIT or Apache 2.0" in the long run.

package_scanner.nim
```diff
@@ -1,135 +1,367 @@
...
+  ]  ## All valid known licences for Nimble packages, on lowercase.
```
dom96: Nitpick: I prefer these to be written as appropriate, i.e. uppercase for acronyms and properly capitalised. You could programmatically pre-process this list and make it all lower-case.

package_scanner.nim
```diff
@@ -1,135 +1,367 @@
...
+            nimble_nonexistent.add url
+
+      # Warn or Assert the possible errors at the end.
+      if nonexistent.len > 0 and allowBrokenUrl:
+        warn "Missing repos list:\n" & nonexistent.join("\n  ")
+        warn "Missing repos count: " & $nonexistent.len & " of " & $pckgsList.len
+      else:
+        doAssert nonexistent.len == 0, "Missing repos: Broken Packages."
+      if nimble_nonexistent.len > 0 and allowMissingNimble:
+        warn "Missing Nimble files:\n" & nimble_nonexistent.join("\n  ")
```
dom96: Nitpick: please stick to camelCase throughout.

package_scanner.nim
juancarlospaco: Its not Silent, it prints a Hint using the Hint Pragma with explicit message.
Note that by default it runs with with SSL then it goes online.
But I get the idea, you want another additional `define` :grey_question:  :thinking:

package_scanner.nim
juancarlospaco: is compile time disabled, if some day want to enable it. :slightly_smiling_face:

package_scanner.nim
juancarlospaco: Why tho :grey_question: :thinking: Is just what the manual says, is something broken with the feature or something :grey_question: 
It wont say experimental nor deprecated.
Not like super important, I will remove it, just curious.

package_scanner.nim
juancarlospaco: I agree, but I am not secure how to implement that, maybe for next PR.

package_scanner.nim
juancarlospaco: u right

.gitignore
```diff
@@ -0,0 +1,4 @@
+*.c
```
FedericoCeratto: Why those ignores?

.travis.yml
```diff
@@ -1,26 +1,30 @@
...
       sh init.sh -y
-  - export PATH=$HOME/.nimble/bin:$PATH
 
 before_script:
   - set -e
   - set -x
 
 script:
-  - nim c -d:ssl -r package_scanner.nim && node ./validate_json.js
+  - nim c -d:ssl -d:release -r package_scanner.nim
```
FedericoCeratto: -d:release slows down the compile time and does not improve significantly the tool run speed

package_scanner.nim
```diff
@@ -1,135 +1,367 @@
...
 
-proc canFetchNimbleRepository(name: string, urlJson: JsonNode): bool =
-  # The fetch is a lie!
-  # TODO: Make this check the actual repo url and check if there is a
-  #       nimble file in it
-  result = true
-  var url: string
+func stripEndingSlash(url): string {.inline.} =
+  ## Strip the ending '/' if any else return the same string.
+  result = if url[^1] == '/': url[0 .. url.len - 2] else: url
```
FedericoCeratto: url[0..^2]

package_scanner.nim
```diff
@@ -1,135 +1,367 @@
...
 
-  if not urlJson.isNil:
-    url = urlJson.str
+func preprocessUrl(url, name): string =
+  ## Take **Normalized** URL & Name return Download link. GitHub & GitLab supported.
+  if likely(url.startswith("https://github.com/")):
```
FedericoCeratto: The tool will spend most time doing networking and little else. Optimizing for branch prediction sounds quite unnecessary :)

package_scanner.nim
```diff
@@ -1,135 +1,367 @@
...
 
-proc verifyAlias(pdata: JsonNode, result: var int) =
-  if not pdata.hasKey("name"):
-    echo "E: missing alias' package name"
-    result.inc()
+proc existsGitRepo(url): string =
+  ## Take **Normalized** URL try to Fetch the Git repo index page. Needs SSL.
+  if url.startswith("http"):
+    try:
+      doAssert client.get(url).status == $Http200 # Check that Repo Exists.
```
FedericoCeratto: Nitpick: in theory setting assertions and catching them is not supposed to be used for control flow and you can easily do if .. raise

package_scanner.nim
```diff
@@ -1,135 +1,367 @@
...
+    result = url  # SSH or other non-HTTP kinda URLs?
 
-  # TODO: Verify that 'alias' points to a known package.
 
-proc check(): int =
-  var name: string
-  echo ""
-  let pkg_list = parseJson(readFile(getCurrentDir() / "packages.json"))
-  var names = initSet[string]()
+suite "Packages consistency testing":
```
FedericoCeratto: Unittest is not meant to test things outside code. It looks like it's testing the tool itself...

package_scanner.nim
```diff
@@ -1,135 +1,367 @@
...
-      elif pdata["url"].str.normalize.startsWith("git://github.com/"):
-        echo "E: ", name, " has an insecure git:// URL instead of https://"
-        result.inc()
+  test "Check file permissions":
+    check getFilePermissions(packagesFilePath) == defaultFilePermissions
+
+  test "Check validate whole JSON by NodeJS":
+    check execCmd(nodeCmd) == 0
+
+  test "Check validate whole JSON by Python":
```
FedericoCeratto: Testing if a JSON file is correct in 3 languages sounds a bit unneded

package_scanner.nim
FedericoCeratto: It's more code to be maintained...

package_scanner.nim
juancarlospaco: What would be the proper wording :grey_question:

package_scanner.nim
juancarlospaco: On the JSON theres weird casing, I always think that doing compare is better on lowercase,
this code is not mean to be seen by end user whatsoever.

Or you mean enforcing proper case on the JSON :grey_question: 
:thinking:

FedericoCeratto:
We are starting to get into the infamous problem of library fragmentation that affects some other languages. I think we should encourage people not to publish tiny libraries. One way is to add a check on codebase size here or in Nimble and politely warn the developer where needed.

juancarlospaco:
I understand the idea about a capitalization linter thing, and the idea about size checking,
but I think those can come on a future PR, the diff is already huge, is working and green,
so maybe merge this and more improvements can come on the future, on other PR.

:slightly_smiling_face:

package_scanner.nim
dom96: I would prefer a run-time argument. Also I want the default behaviour to be "online".

package_scanner.nim
dom96: > this code is not mean to be seen by end user whatsoever.

We are the end users. We use this scanner to tell us whether the newly added package is valid, this is just a style choice though so it's not a big deal. I don't see what JSON has to do with this.

package_scanner.nim
juancarlospaco: The default behavior to be "online".
No extra option required anywhere.
All other is for when a dev want to run it manually for whatever reason.

:slightly_smiling_face:

juancarlospaco:
@dom96 Let me know any other feedback.

package_scanner.nim
dom96: No, the default is ``offline``. Default means "compiling with no options".

dom96:
We still have unresolved feedback here. Please comment on each why you disagree with me or fix it.

juancarlospaco:
@dom96 Check again, let me know if you see any problem.

juancarlospaco:
@dom96 Let me know if something is still missing...
:thinking:

xflywind:
I closed it because https://github.com/nim-lang/packages/pull/2027 is merged.

PR #917, opened 2018-10-31
Add nimetry module

ijneb:


dom96:
Huh, interesting. I got an error the first time I attempted to install this:

```
Error: Cannot satisfy the dependency on vmath 0.2.1 and vmath 0.1.0
```

Guess this is a Nimble bug.

PR #919, opened 2018-11-03
Added package nimzbar

genotrance:


dom96:
Wow. Nimble crashes with no error when installing this package.

genotrance:
I'll bet you are testing on Mac which I couldn't get working since there is no makefile I could refer to. I've put in an exit on Mac with a message and updated Readme but nimble doesn't print anything. It does work on other OS though.

dom96:
> https://irclogs.nim-lang.org/15-11-2018.html#00:38:41

I'm not going to add this package. Quitting with no message is terrible during an install.

genotrance:
Looks like I missed an echo - I had one in nimFastText. I've pushed a fix with a message.

https://github.com/genotrance/nimzbar/blob/master/nimzbar.nimble#L19

However, even with an echo, nimble doesn't print anything, but quits for sure.

dom96:
Thanks!

PR #934, opened 2018-11-18
Add package nish

owlinux1000:


Yardanico:
@owlinux1000 it seems that you've opened a PR two times - please close either this one or https://github.com/nim-lang/packages/pull/933 :)

owlinux1000:
I'm sorry for PR duplicated. 
This is the first time using nimble publish.
I'll close #933!

thanks

packages.json
```diff
@@ -9295,7 +9295,7 @@
       "pqcrypto",
       "signing"
     ],
-    "description": "SPHINCS⁺ stateless hash-based signature scheme",
+    "description": "SPHINCS\u207A stateless hash-based signature scheme",
```
treeform: I think unicode crap happened?

dom96: Not @owlinux1000's fault and it doesn't matter too much IMO.

PR #942, opened 2018-11-21
Add package genode

ehmry:


ehmry:
This package is for OS-specific APIs as well as configuring the Nim compliler with `passC` and `passL` pragmas. This is better implemented in a Nimble package that increments with each Genode release so that the main Nim repository not be tied to a specific release.

PR #943, opened 2018-11-25
Add bgfxdotnim

zacharycarter:
Add bgfxdotnim (BGFX C99 bindings)

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -5476,6 +5476,22 @@
     "method": "git"
   },
   {
+    "name": "bgfx",
+    "url": "https://github.com/Halsys/nim-bgfx",
+    "method": "git",
+    "tags": [
+      "wrapper",
+      "media",
+      "graphics",
+      "3d",
+      "rendering",
+      "opengl"
+    ],
+    "description": "BGFX wrapper for the nim programming language.",
+    "license": "BSD2",
+    "web": "https://github.com/Halsys/nim-bgfx"
+  },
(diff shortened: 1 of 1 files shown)


narimiran:
1. @dom96 how do you feel about removing a package from nimble?

2. @zacharycarter Before merging, I would recommend adding `README.md` to your repository, where you explain what your package does, and how to use it.

zacharycarter:
@narimiran - will add a README this morning.

Regarding the existing BGFX package in nimble's dir - I can certainly amend the PR to keep the existing BGFX bindings from halsyss but since they haven't been updated in over a year and they're not usable with the current version of BGFX, I wonder if there's any point in keeping them in the nimble package repo.

Araq:
@narimiran A Nimble "package repository" where stuff can disappear from nilly-willy is **unusable** IMHO. Nimble needs to fork every registered nimble package and old stuff should live forever. Exception: Packages with terrible security implications.

narimiran:
@Araq I thought that might be the case. Now I closed #944 for the same reason.

@zacharycarter will you modify this PR to only add your package?

zacharycarter:
Will do, thanks!

dom96:
Yes, what @Araq said. Packages **should not** be removed, for any reason. This is one of the reasons why we need our own centralised package repo.

zacharycarter:
@narimiran done!

PR #944, opened 2018-11-25
Removing frag from package list

zacharycarter:
This package is no longer useful - proposing a PR to remove it from nimble.


narimiran:
Closed because of https://github.com/nim-lang/packages/pull/943#issuecomment-441706588

PR #955, opened 2018-12-05
Added package nimfastText

genotrance:


narimiran:
```
Downloading https://github.com/genotrance/nimfastText using git
../../../compiler/lineinfos.nim(228) raiseRecoverableError
Error: unhandled exception: unhandled exception: Install failed, package not supported on gcc < 5.x [ERecoverableError]
```

genotrance:
Yes the package isn't supported with older gcc since it uses latest c++ standard.

Message is raised as requested by @dom96.

dom96:
That's a real shame. Does it also not support clang at all?

genotrance:
nimfastText works fine on OSX/clang actually but the gcc check was happening everywhere and blocking on OSX.

I'm updating the nimble file to check for gcc version only on Windows/Linux. This should now work on OSX.

dom96:
It just quits with no message now.

dom96:
Wouldn't it be simpler to just check this at compile-time vs. at install time? That way users can go in and edit the code if they want to force it to work

genotrance:
> It just quits with no message now.

That's because it is supported on MacOSX which you are probably trying. It isn't supported on Linux on Travis for example which uses an old gcc. It works everywhere with a modern compiler.

> Wouldn't it be simpler to just check this at compile-time vs. at install time? That way users can go in and edit the code if they want to force it to work

This is to inform users about a tool dependency which took me some time to figure out. Rather than wasting peoples' time, they get a clear message telling them:

`Install failed, package not supported on gcc < 5.x`

They can always update their gcc and try again.

genotrance:
I'm merging this. If we need any updates, I'll be happy to update the package repo.

PR #957, opened 2018-12-07
Add tree-sitter and nimterop

genotrance:


dom96:
> These are separate packages because each language is ~10MB worth of source code that has to be downloaded.

How are these downloaded?

> Meanwhile, even if installed separately, they are all imported using treesitter/X so the import path is consistent despite being different packages.

How does that work? Your package's are in an `treesitter_X` dir.

genotrance:
> How are these downloaded?

They are downloaded by nimgen using git. The 7-10MB repo size per language is a sparse checkout and the C/C++ source files are large since they are auto-generated by the tree-sitter compiler.

> How does that work? Your packages are in an `treesitter_X` dir.

I was doubtful myself but it works - here's the directory structure and the Nim compiler resolves the directories just fine.

```
~/.nimble/pkgs/treesitter-0.1.0/treesitter/runtime.nim
~/.nimble/pkgs/treesitter_c-0.1.0/treesitter/c.nim
~/.nimble/pkgs/treesitter_cpp-0.1.0/treesitter/cpp.nim
```

```nim
import treesitter/[runtime, c, cpp]
```

genotrance:
The h files are useless by themselves, source is required and compiled in.

I can change the namespaces to be unique but it is unnecessary for related packages. Nimble has more than enough rules for package maintainers.

dom96:
I would love to accept this, and I can do that, but your packages will eventually stop working because they break Nimble's namespace rules. I'd rather new packages followed these rules.

genotrance:
We can discuss this topic at length and figure out best steps forward but that shouldn't gate entry into the nimble directory. The packages work today and will continue to be maintained.

After contributing over 25 packages in the last two years, all of which are maintained and still functional across multiple Nim releases, I think there should be sufficient credibility to my efforts. If I have to wait weeks and resolve every single concern before you accept anything then I'd rather spend my efforts elsewhere.

genotrance:
These packages work just fine today despite flouting these namespace rules - the rules are not enforced nor documented. I understand the principal but not sure how you can enforce this programmatically without testing every single package simultaneously and verifying that they don't override each other. Meanwhile, I've raised a use case where it perhaps makes sense to do this but I am also receptive to changing it since `import treesitter_c/c` vs. `import treesitter/c` is of little practical consequence.

I understand this is to prevent packages from colliding but I'm not sure how that can be achieved.

dom96:
> the rules are not enforced nor documented.

They are. Nimble prints warnings when packages don't conform and these rules are documented in the Nimble readme.

I will accept this, but please change it to ``treesitter_c`` or merge everything into a single ``treesitter` package (I would much prefer that). You could also do some hacks and provide a "forwarding module" inside your ``treesitter`` package that exports ``treesitter_c`` etc.

dom96:
I'm not 100% sure I understand your question. The long-term goal is to get the CI for this repo to test by installing the package.

genotrance:
Okay, that's what I was trying to understand.

I've fixed the treesitter namespace issue you raised. For now, I'm using `import treesitter_c/c` in consumers. However, I'm not sure how to forward a module after checking if it is installed or not.

E.g.
```nim
when installed(treesitter_c):
  import treesitter_c/c
  export treesitter_c/c
```

dom96:
Maybe:

```nim
when compiles(import treesitter_c/c): ...
```

or something like that

PR #964, opened 2018-12-12
adding whip server

mattaylor:


narimiran:
When I try to install it, I get the following warning:

   Warning: Package 'whip' has an incorrect structure. The top level of the package source directory should contain at most one module, named 'whip.nim', but a file named 'tests.nim' was found. This will be an error in the future.
      Hint: If this is the primary source file in the package, rename it to 'whip.nim'. 
If it's a source file required by the main module, or if it is one of several modules exposed by 'whip', then move it into a 'whip/' subdirectory. 
If it's a test file or otherwise not required to build the the package 'whip.nim', prevent its installation by adding `skipFiles = @["tests.nim"]` to the .nimble file. 
See https://github.com/nim-lang/nimble#libraries formore info.


Would you please solve this issue before we merge this, so your users don't see the same message when they install it?

mattaylor:
should be fixed now - please try again

PR #968, opened 2018-12-19
variantkey

brentp:


brentp:
@narimiran I could be mistaken, but I think that is a nim/nimble issue as I have been getting that error with various packages lately when using devel. Not sure why or how I resolved it.
If it's something I've done in the nimble file, I'll be glad to adjust it, but I don't see any problems and "it works for me".

SolitudeSF:
works for me on the latest devel. @narimiran, could you try with nimble compiled with latest compiler?

PR #977, opened 2019-01-04
Added package wren

geotre:


genotrance:
This is a nimble issue where dependencies aren't installed before the `before install` pre hook.

See my [workaround] (https://github.com/genotrance/nimarchive/blob/master/nimarchive.nimble). 

Nimble issue is [here](https://github.com/nim-lang/nimble/issues/533).

geotre:
@genotrance thanks for the tip, I've copied the method you used! I saw that when I looked at your packages before but didn't understand why it was needed so didn't use it.

@narimiran that is fixed now, could you try again?

narimiran:
```
    Error: Unsatisfied dependency: nim (>= 0.19.9)
```

Is your package really dependent on the latest devel and it doesn't work with the stable version (v0.19.2) of Nim? If so, IMO it shouldn't be merged.

If you're using `choosenim`, please do `choosenim stable` and test if your package works with the stable version. If it does, please correct the dependency in your `.nimble` file.

(More descriptive `README.md` would also be nice)

geotre:
@narimiran you are correct, I think `nimble init` took the current running version of nim for the dependency. I've tested on stable and updated the nimble file.

I've also updated the readme to look more like the other nimgen generated packages

PR #995, opened 2019-01-21
Add package ladder

ryukoposting:


ryukoposting:
I'm hosting the generated docs [here.](http://ryuk.ooo/nimdocs/ladder/ladder.html)

PR #1000, opened 2019-01-31
Add package lastfm

ryukoposting:


evanperryg:
hosting documentation here: http://ryuk.ooo/nimdocs/lastfm/lastfm.html

PR #1013, opened 2019-02-22
remove readme link to Nim "library overview"

citycide:
The referenced section was removed from Nim's documentation.


narimiran:
Maybe we should change that paragraph so now it links to https://nimble.directory/ ? @dom96?

PR #1021, opened 2019-03-01
Deprecate Gatabase

juancarlospaco:
Deprecate Gatabase.


Araq:
This is something that a Nimble package repository shouldn't support. Stuff shouldn't disappear under one's feet.

PR #1024, opened 2019-03-06
Add package tmpgen

kekeho:


narimiran:
I wasn't able to test-install it via `nimble install https://github.com/kekeho/tmpgen` on Nim 0.19.9 :/

kekeho:
sorry for late reply.
I see. I'm beginner of nim, so I'll try fix it.
thanks

PR #1027, opened 2019-03-12
Add libp2p

zah:


dom96:
ooh, did you guys wrap BearSSL? Can we get a nimble package for that?

zah:
We did, I'll see what I can do.

cheatfate:
@dom96 its not full wrap of BearSSL. But here still present https://github.com/Araq/bearssl

PR #1044, opened 2019-03-28
Marc azar patch 1

MarcAzar:


dom96:
@narimiran Unlikely to matter for this package but please keep this in mind: https://github.com/nim-lang/packages#renaming-packages

PR #1045, opened 2019-03-29
Add package nim_curry

zer0-star:


narimiran:
Before I merge it, can you write a short description and/or an example in your readme what your package does and how can it be used?

zer0-star:
ok, i wrote.

PR #1057, opened 2019-04-13
Add package gameoflife

jiro4989:


jiro4989:
Oh, sorry. URL is old. I will fix URL.

PR #1067, opened 2019-04-18
Add package jsutils

kidandcat:


SolitudeSF:
tags property is for tags describing the package, not the git tags.

kidandcat:
Sorry, confused, changing it!

PR #1098, opened 2019-06-20
Updating the bignum package source to a fork that works on Nim 0.20

kaushalmodi:
The original repo https://github.com/FedeOmoto/bignum hasn't been
maintained in many years.

I am not an expert in bignum but I am maintaining a fork of it that
runs with the latest stable Nim version (v0.20.0 as of today).

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -2989,7 +2989,7 @@
   },
   {
     "name": "bignum",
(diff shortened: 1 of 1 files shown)


dom96:
CC @FedeOmoto

PR #1101, opened 2019-06-25
Add package dnsclient

ba0f3:


narimiran:
Before merging, can you write a short readme where you explain what your package is, what does it do, and how to use it?

ba0f3:
done

PR #1112, opened 2019-07-09
Add optionsutils package

PMunch:


trustable-code:
URLs are wrong. Probably should be https://github.com/PMunch/nim-optionsutils.

PMunch:
Woops, fixed now

PR #1116, opened 2019-07-13
Add package illwill

johnnovak:


narimiran:
See my comment on your previous PR — it is already merged (see the lines above your addition).

From what I can tell, in this PR you only need to change the license, no need for additional entry.

PR #1118, opened 2019-07-15
Add totp and hotp

CORDEA:


ba0f3:
conflict w/ my [otp](https://github.com/OpenSystemsLab/otp.nim/blob/master/otp.nim) moudle

CORDEA:
Oops, thank you @ba0f3.

PR #1121, opened 2019-07-22
Add package fastx_reader

ahcm:


narimiran:
While we wait for CI to complete its job, a more detailed readme would be nice ;) What does this package do, how to use it?

PR #1129, opened 2019-07-25
Add package MQTT

kidandcat:
Add package MQTT


narimiran:
The last commit was 4 years ago: are we sure this still compiles/works with the current Nim?

kidandcat:
Yes, I use it with 0.20 on windows

PR #1144, opened 2019-08-19
Change twitter package maintainer

dchem:
This patch changes the twitter package maintainer. @kubo39 is the original developer, but has decided to archive the project, so no further pull request can be made.

The forked version `0.2.1` fixes the build issues due to changes in latest nim.


dom96:
Thank you for taking over the maintenance of this!

PR #1165, opened 2019-09-13
Extended metadata information

pmetras:
This pull request adds new optional metadata fields to a package description:
* `long_description`: a multi-lines description that can be used for package searches.
* `code_quality`: a code quality indicator in the range 1 (low) to 4 (high).
* `doc_quality`: a documentation quality indicator in the range 1 (low) to 4 (high).
* `project_quality`: a project vitality indicator in the range 0 (dead) to 4 (high community).
* `categories`: static categories
* various other metadata: see documentation https://github.com/pmetras/packages/blob/extended_info/README.md...

All the packages in `packages.json` have been curated to find these new metadata from the repositories, when easily available. The full list of packages that has been used to generate `package.json` is available in https://docs.google.com/spreadsheets/d/1HWy2YumMMcgEDHk34ACauuWR5TYDJTRUVQ6B-LuRXCs/edit?usp=sharing, extracted from Github on 2019-07-27 and updated with new packages up to today (2019-09-12).

This extra information is used by the features added to `nimble` in pull request https://github.com/nim-lang/nimble/pull/691.

As `packages.json` is generated from a Google Sheet where people changed the sort order, the file is no more chronological and I've sorted it alphabetically. As long as we keep it sorted, `nimble --list` gives a nice sorted output :-), but it can't be compared any more with a previous version in another branch ;-((. You have to trust me that I did not forgot a few packages...

Another problem is that I did the error to push binary executables into my branch and I had to purge git history from these files. Now the branch is thousands commit ahead of master! Tell me if you prefer that I start back from master and lose my history instead.

As I did the packages list curation alone, perhaps packages authors won't agree with the metadata I selected. I'm going to publish a call to package authors on Nim Forum to give them the opportunity of changing the metadata in the Google Sheet. When completed, I'll be able to regenerate the final version of `packages.json`. You can use this amount of time to test and see if this extra metadata information is worth adding...


treeform:
I don't think `*_quality` metrics should be part of this json file. They are very subjective and might discourage other from making packages when official repo marks them as "bad".

I agree with other things mostly.

dom96:
IMO this data should be stored in a separate location. I could have sworn I wrote this down somewhere else...

PR #1186, opened 2019-09-27
Add package wings

binhonglee:


Araq:
Why close this? It looked nice!

binhonglee:
@Araq, the package is already published previously. I just got confused by the flow to update the package. Should probably add some sort of checks before `nimble publish` get fired off.

PR #1190, opened 2019-09-28
Add package the_nim_alliance

tervay:


narimiran:
Can you please write a README where you describe your package: what does it do and how to use it?

tervay:
Added; sorry about that, first Nim package, still figuring things out.

PR #1192, opened 2019-09-29
Add package PPM

LemonHX:


LemonHX:
then i need to remove bin in nimble right？

dom96:
Yes or follow https://github.com/nim-lang/nimble#hybrids

LemonHX:
@dom96 # Fixed

PR #1197, opened 2019-10-07
Added bearssl package

yglukhov:
Added bearssl package. Lost of copypaste from Araq's lib with small fixes. Hope you don't mind @Araq.


Araq:
I never mind people making good use of my code.

PR #1205, opened 2019-10-15
Add package rtree

StefanSalewski:


packages.json
```diff
@@ -14788,5 +14788,23 @@
...
+  },
+  {
+    "name": "rtree",
+    "url": "https://github.com/stefansalewski/RTree",
+    "method": "git",
+    "tags": [
+      "RTree",
+      "R-Tree",
+      "R*Tree",
+      "R*-Tree",
```
dom96: Please remove these. These aren't good tags, they are just variations of the name. Tags should reflect some overarching categories like `library` or `wrapper` or `network` etc.

StefanSalewski:
OK -- I clicked on something like "Resolve conversation".

But package seems to be not in database still. Have I to call nimble publish again, with shorter tags list?

No time to further investigations currently, sorry.

narimiran:
> OK -- I clicked on something like "Resolve conversation".

That just hid the conversation, nothing else changed.

StefanSalewski:
Seems to be not that easy, from

https://stackoverflow.com/questions/16748115/how-to-modify-github-pull-request

my feeling is that I have to learn a lot of git/github API. I will try to close this one and do a new nimble publish.

PR #1227, opened 2019-10-29
Added package autotyper

kijowski:


kijowski:
Thanks for catching that bug - it should be fixed now

PR #1233, opened 2019-11-01
Add package norm

jonasrauber:


narimiran:
Package of that name already exists. You'll have to think of another name.

PR #1235, opened 2019-11-03
Remove packages with dead source URLs

ehmry:


ehmry:
Thats everything I could find for now.

PR #1238, opened 2019-11-05
Add package cli_menu

MnlPhlp:


narimiran:
Some examples in the readme would be nice to have.

PR #1242, opened 2019-11-09
Add package bson

JohnAD:
Added package bson


JohnAD:
My bad. I added two in a row and it ended up in the same PR. Will cancel and try again.

PR #1264, opened 2019-11-30
Add nimnoise package

blakeanedved:


blakeanedved:
My bad, the file permissions should be unchanged now.

PR #1274, opened 2019-12-16
rename evmjit to evmc

jangko:
jit-ed or not, the correct name should be "evmc".


jangko:
> Has this name ever been correct?

it has been not maintained for 2 years, we just revived it.

PR #1279, opened 2019-12-18
Add package diff

mark-summerfield:


packages.json
```diff
@@ -15675,5 +15675,17 @@
     "description": "This library is a wrapper to C GDBM library",
     "license": "MIT",
     "web": "https://github.com/vycb/gdbmc.nim"
+  },
+  {
+    "name": "diff",
+    "url": "https://github.com/mark-summerfield/packages",
```
narimiran: This is not a correct link to your package.

mark-summerfield: I just did ``nimble publish`` and answered as best I could. This is the first time I've tried it. What should the url be then?

SolitudeSF: url to your package?

mark-summerfield: I thought that ``nimble publish`` created the package and put it on github?
If not, I presume I must somehow get it to github but although I have a github a/c I've never uploaded before, so I don't know how to do that. Presumably, then I'd give a github url?
Or can I just put it on my website? But there doesn't seem to be a nimble command for creating a package?

PR #1288, opened 2019-12-24
Add package rfc3339

Skrylar:


packages.json
```diff
@@ -15687,5 +15687,16 @@
     "description": "Library for finding the differences between two sequences",
     "license": "Apache-2.0",
     "web": "https://github.com/mark-summerfield/diff"
+  },
+  {
+    "name": "rfc3339",
+    "url": "https://git.sr.ht/~skrylar/skrfc3339",
+    "method": "git",
+    "tags": [
+      "rfc3339,date,datetime,time"
```
dom96: Each tag should be its own item

Skrylar:
That was all generated with nimble /shrug

dom96:
Yes, Nimble should detect this. But it's instructions are clear: separate tags with spaces not commas.

narimiran:
> `E: rfc3339: a package by that name already exists.`

...and it is your package hosted on Github: https://github.com/Skrylar/rfc3339

Skrylar:
Was just trying to do a version bump. Doesn't look like this is necessary anyhow.

PR #1290, opened 2019-12-26
Add validator string validation library

Adeohluwa:

PR #1287, opened 2019-12-24
Add package simhash

bung87:

PR #1283, opened 2019-12-22
Add package lq

madprops:

PR #1281, opened 2019-12-21
Add package taps

ehmry:

PR #1291, opened 2019-12-26
add multiple packages

narimiran:


narimiran:
This adds packages from #1281, #1282, #1283, #1284, #1285, #1286, #1287, #1290

PR #1346, opened 2020-01-29
Add package react

kristianmandrup:


narimiran:
This won't do, because, as Travis reports: `E: react: a package by that name already exists.`

Secondly, you've already added your package in https://github.com/nim-lang/packages/pull/1345

PR #1354, opened 2020-02-11
Add package shizuka

Ethosa:


SolitudeSF:
you should also probably add that your previous library is deprecated.

PR #1359, opened 2020-02-17
Add package nYara

bartimus-primed:


bartimus-primed:
Ill add this once I get the library done.

PR #1382, opened 2020-03-02
Add package nimbus

paul-nameless:


Clyybber:
There is already https://github.com/status-im/nimbus so this would conflict :/

PR #1401, opened 2020-03-16
Add package sss

markspanbroek:


packages.json
```diff
@@ -16989,5 +16989,18 @@
     "description": "Native MQTT client library",
     "license": "MIT",
     "web": "https://github.com/zevv/nmqtt"
+  },
+  {
+    "name": "sss",
+    "url": "https:///",
+    "method": "",
```
narimiran: You need fill the `method`, and other fields too.

markspanbroek: Thanks @narimiran, I used 'nimble publish' and didn't see that it failed to fill in some fields. I'll update them by hand in a minute.

PR #1415, opened 2020-03-27
Add package nimcoon

JosephKiranBabu:
NimCoon is a thin wrapper over youtube-dl, mpv and peerflix that allows you to stream or download music and video from sites like YouTube, PeerTube and magnet links.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -17162,5 +17162,20 @@
     "description": "libvlc bindings for Nim",
     "license": "MIT",
     "web": "https://github.com/Yardanico/nim-libvlc"
+  },
+  {
+    "name": "nimcoon",
+    "url": "https://njoseph.me/gitweb/nimcoon",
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -17162,5 +17162,20 @@
     "description": "libvlc bindings for Nim",
     "license": "MIT",
     "web": "https://github.com/Yardanico/nim-libvlc"
+  },
+  {
+    "name": "nimcoon",
+    "url": "https://njoseph.me/gitweb/nimcoon",
```
narimiran: I might be wrong, but this doesn't look like a correct url for this package; it lists all of your projects.

SolitudeSF: git clones its fine

JosephKiranBabu: Adding .git to the end of the url takes to the correct page. I will make the correction.

PR #1441, opened 2020-04-22
added nimqt

faulander:
# nimtm - a wrapper for the QBittorrent Web-API
## QUICK EXAMPLE
```nim
import nimtm
let connection = initQb("http://127.0.0.1:8080", "admin", "adminadmin")
conn.login()
if conn.connected:
  var torrents = conn.getTorrents()
  for torrent in torrents:
    echo getTorrentFiles(torrent["hash"])
```

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -17463,5 +17463,19 @@
     "license": "zlib",
     "web": "https://github.com/exelotl/trick",
     "doc": "https://exelotl.github.io/trick/trick.html"
+  },
+    {
+    "name": "nimqt",
+    "url": "https://github.com/faulander/nimqt",
+    "method": "git",
+    "tags": [
+      "torrent",
+      "qbittorrent",
+      "api",
+      "wrapper",
+    ],
+    "description": "a wrapper for the QBittorrent WebAPI for NIM.",
+    "license": "MIT",
+    "web": "https://github.com/faulander/nimqt"
   }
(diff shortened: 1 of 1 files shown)


Yardanico:
For a second I thought that this was a Qt wrapper by the package name

packages.json
```diff
@@ -17463,5 +17463,19 @@
...
+  },
+    {
+    "name": "nimqt",
+    "url": "https://github.com/faulander/nimqt",
+    "method": "git",
+    "tags": [
+      "torrent",
+      "qbittorrent",
+      "api",
+      "wrapper",
```
narimiran, line 17475: You need to remove this last comma (`,`), otherwise it is invalid JSON.

faulander:
> For a second I thought that this was a Qt wrapper by the package name

I am sorry, but that's the appreviation they use for QBittorrent, i would have named it Qb otherwise. But i can rename it?

narimiran:
> [qt is] the abbreviation they use for QBittorrent

You sure about that?

> But i can rename it?

I think lots of people would be confused to see that `qt` means `qbittorrent`, because they expect that it stands for... well... [Qt](https://en.wikipedia.org/wiki/Qt_(software)) :)

faulander:
renamed nimqt to nimQBittorrent

PR #1463, opened 2020-05-08
Add package bstreams

sealmove:


packages.json
```diff
@@ -17768,5 +17768,16 @@
     "description": "An alternative time library",
     "license": "MIT",
     "web": "https://github.com/jackhftang/timestamp.nim"
+  },
+  {
+    "name": "bstreams",
+    "url": "https://github.com/sealmove/bstreams",
+    "method": "git",
+    "tags": [
+      "library"
```
narimiran, line 17777: It would be beneficial if you added more tags, otherwise your package will be very hard to search and to find it.

(Also, it needs a more informative `readme`, where you explain what that package is, what it does, and how to use it)

sealmove: You are right, it's actually a premature publication. Please don't merge this yet, it might end up not worth merging. Thanks for the review!

PR #1471, opened 2020-05-11
Add package stalinsort

Knaque:


Yardanico:
But does it send bad numbers to Gulag?

Knaque:
It `delete`s them, so... Close enough, as far as the Soviets would be concerned.

PR #1479, opened 2020-05-17
Added package buffer

bung87:


packages.json
```diff
@@ -17906,5 +17906,18 @@
...
     "web": "https://github.com/xzeshen/bio"
+  },
+  {
+    "name": "buffer",
+    "url": "https://github.com/bung87/buffer",
+    "method": "git",
+    "tags": [
+      "stream",
+      "buffer",
+      ""
```
narimiran, line 17917: An empty tag? Can you remove it?

bung87: ok, did not aware of it

PR #1480, opened 2020-05-17
Add package events

al-bimani:


packages.json
```diff
@@ -17919,5 +17919,20 @@
     "description": "Desktop notifications",
     "license": "MIT",
     "web": "https://github.com/SolitudeSF/notification"
+  },
+  {
+    "name": "events",
+    "url": "https:///",
```
Yardanico, line 17925: Where's the actual url? :)

PR #1483, opened 2020-05-21
Add package nmi

jiro4989:


packages.json
```diff
@@ -17957,5 +17957,18 @@
...
+    "url": "ssh://git@github.com/jiro4989/nmi",
+    "method": "git",
+    "tags": [
+      "sl",
+      "joke",
+      "cli"
+    ],
+    "description": "nmi display animations aimed to correct users who accidentally enter nmi instead of nim.",
+    "license": "MIT",
+    "web": "ssh://git@github.com/jiro4989/nmi"
```
Clyybber, line 17972: This should not be an ssh link

jiro4989: oh sorry

jiro4989: Fixed

PR #1487, opened 2020-05-23
add package taskqueue and threadproxy

jackhftang:
add doc of one of my package

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -1,4 +1,5 @@
-[{
+[
+  {
     "name": "filesize",
     "url": "https://github.com/sergiotapia/filesize",
     "method": "git",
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -18012,7 +18012,7 @@
     "license": "MIT",
     "web": "https://github.com/Ethosa/nodesnim"
   },
-  {
+  {
```
Clyybber, line 18015:
```suggestion
  {
```
only suggesting this for myself :)

packages.json
```diff
@@ -1,5 +1,4 @@
-[
-  {
+[{
```
Clyybber, line 1:
```suggestion
[
  {
```
only suggesting this for myself :)

PR #1491, opened 2020-05-28
Add package jesterwithplugins

JohnAD:
By arrangment with dom96, this package was forked from `jester`. About a dozen or more "plugin" packages will be added that take advantage of this `jester` fork. Later (6 months to a year), after all lessons are learned :) and the code is merged up, we will deprecate this package.

I'll add the first four plugin packages after this goes live.


narimiran:
Added in #1500.

PR #1495, opened 2020-05-30
Add package qeu

hyu1996:
Functionality for compare two values.
Supports comparison between signed and unsigned integers.


Yardanico:
Just for your information - https://nim-lang.org/docs/lenientops.html exists to compare different number types

narimiran:
Added in #1500.

PR #1505, opened 2020-06-08
add package vaultclient

jackhftang:


packages.json
```diff
@@ -1,5 +1,4 @@
-[
-  {
+[{
```
Yardanico, line 1: Please don't make changes to the JSON itself :)

jackhftang: Oh, it is vscode default format-on-save. I thought it will be auto re-format again when merge =p

PR #1510, opened 2020-06-11
Add package whip

mattaylor:


Yardanico:
I think your diff is a bit messed up :)

narimiran:
Travis CI reports the following error:

`whip: a package by that name already exists`

PR #1525, opened 2020-06-20
Add package nim_searches

nnahito:


narimiran:
Changing descriptions of other packages is a no-no.

nnahito:
I just ran nimble publish, so I thought it was normal.
I'll fix it.

PR #1531, opened 2020-06-23
Add package nim_searches

nnahito:


Yardanico:
How are you making PRs? :) You can just use `nimble publish` to make sure that everything works correctly and you don't change other descriptions. Or maybe your Nim version is outdated?

PR #1538, opened 2020-06-27
Add package soundex

Kashiwara0205:


packages.json
```diff
@@ -18504,5 +18504,17 @@
     "description": "Nim CMAES library",
     "license": "Apache-2.0",
     "web": "https://github.com/zevv/nimcmaes"
+  },
+  {
+    "name": "soundex",
+    "url": "https:///",
```
Yardanico, line 18510: That doesn't look correct for me :P

Kashiwara0205: Thank you for the review.
I fix it and submit again.

Kashiwara0205: I checked about URL setting but I don't know how to set the URL.
How should I set the URL.

If it's not too much trouble, can you tell me?

Kashiwara0205:
Ｉfix package.json

PR #1541, opened 2020-06-28
Add package vds

doongjohn:
Should I rename this to something else?

Because I found this:


narimiran:
> Should I rename this to something else?

If you ask me, it is up to you.

What the others are saying? @Clyybber @Yardanico ?

doongjohn:
I guess I can change it later if there is a problem.

Yardanico:
I think it's okay - https://docs.microsoft.com/en-us/windows/win32/vds/about-vds is a service and if Nimble adds binary folder to PATH after the system PATHs so it'll work just fine.

PR #1568, opened 2020-07-23
Change twitter.nim owner

snus-kin:
The current owner did not respond to my PR to add support for uploading media to twitter for 30 days and I have added a bunch of procs for doing this, his bio also says 'not hosting on github'

I don't really want to step on anyone's toes here but it seems like the default 'twitter' package should at least support requests to the media endpoint.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -5075,7 +5075,7 @@
   },
   {
     "name": "twitter",
-    "url": "https://github.com/dchem/twitter.nim",
+    "url": "https://github.com/snus-kin/twitter.nim",
     "method": "git",
     "tags": [
       "library",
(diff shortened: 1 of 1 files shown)


narimiran:
@dom96 opinion?

dom96:
Seems reasonable. If @dchem wants it changed back we should allow it though.

dom96:
Might want to give him a ping on dchem@gitlord.com

snus-kin:
Yeah as I said not wanting to step on toes but the library is missing what I would consider core functionality for something that is called 'twitter' in the repo. I'd quite like to see it even if it's not my code in the repo (have been my forked version for: https://github.com/snus-kin/flickr-image-bot).

narimiran:
> Seems reasonable. If @dchem wants it changed back we should allow it though.

Merging. Will revert if necessary (doesn't seem likely).

dchem:
Apologies for late reply. I have been unfortunately focused on other projects. No objections. Thank you.

PR #1569, opened 2020-07-24
Add ms

fox-cat:
(/・・)ノ" a sorta of small library for converting various time formats to milliseconds


narimiran:
There's a problem with our CIs, they are stuck.

This happens from time to time, and the only cure so far has been closing a PR and opening another one. Please open another pull request, maybe it will be more luck next time.

PR #1590, opened 2020-08-09
Add a simple bitarray library

YesDrX:


ba0f3:
[binary](https://github.com/YesDrX/bitarray) should removed from the repo

YesDrX:
> [binary](https://github.com/YesDrX/bitarray/blob/master/src/bitarray) should removed from the repo

removed

YesDrX:
> [This file](https://github.com/YesDrX/bitarray/blob/master/src/utils.nim) will be installed into the global namespace, i.e. it will be importable via `import utils`. You need to put it into a folder named `bitarray`

done.

narimiran:
Merged as a part of #1617.

dom96:
The name of this package is wrong, I had to revert its addition: https://github.com/nim-lang/packages/commit/5416690d7331db79dd6686c35773d895da8b9864. You need to name it something else as `bitarray` is an existing package.

ba0f3:
next time you should make a branch before do pull request

YesDrX:
> next time you should make a branch before do pull request

sure.

YesDrX:
nim-bitarray changed to nimbitarray.

PR #1610, opened 2020-08-21
Add cmark package

zengxs:
[libcmark](https://github.com/commonmark/cmark) wrapper for Nim.


narimiran:
Merged as a part of #1617.

PR #1617, opened 2020-08-24
add multiple packages at once to prevent merge conflicts

narimiran:


packages.json
```diff
@@ -18967,6 +18967,169 @@
...
+      "Simple",
+      "HTTP",
+      "Server"
+    ],
+    "description": "SimpleHTTPServer module based on net sockets",
+    "license": "HYDRA",
+    "web": "https://github.com/Hydra820/NimpleHTTPServer"
+  },
+  {
+    "name": "nim-bitarray",
```
dom96, line 18984: this is wrong, it should be `bitarray`. Dashes should generally not be allowed in package names.

PR #1637, opened 2020-09-11
Added package Blackvas

momeemt:


packages.json
```diff
@@ -19532,5 +19532,21 @@
     "description": "Once provides a type that will enforce that a callback is invoked only once.",
     "license": "BSD3",
     "web": "https://git.sr.ht/~euantorano/once.nim"
+  },
+  {
+    "name": "Blackvas",
+    "url": "https://github.com/momeemt/HighCanvas",
```
narimiran, line 19538: Shouldn't this url be `https://github.com/momeemt/Blackvas`?

momeemt: Yes, I mistake it.

PR #1648, opened 2020-09-29
Added packets

Q-Master:
pure Nim declarative packes system for serializing/deserializing

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -19631,5 +19631,18 @@
     "description": "A fast and simple command line argument parser inspired by Python's argparse.",
     "license": "MIT",
     "web": "https://github.com/oliversandli/clapfn"
+  },
+  {
+    "name": "packets",
+    "url": "https://github.com/Q-Master/packets.nim",
+    "method": "git",
+    "tags": [
+      "serializtion",
+      "deserialization",
+      "marshal"
+    ],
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -19631,5 +19631,18 @@
...
+    "url": "https://github.com/Q-Master/packets.nim.git",
+    "method": "git",
+    "tags": [
+      "serializtion",
+      "deserialization",
+      "marshal"
+    ],
+    "description": "Declarative packets system for serializing/deserializing and marshalling",
+    "license": "MIT",
+    "web": "https://github.com/Q-Master/packets.nim.git"
```
Yardanico, line 19646: You should remove .git there

packages.json
```diff
@@ -19631,5 +19631,18 @@
     "description": "A fast and simple command line argument parser inspired by Python's argparse.",
     "license": "MIT",
     "web": "https://github.com/oliversandli/clapfn"
+  },
+  {
+    "name": "packets",
+    "url": "https://github.com/Q-Master/packets.nim.git",
```
Yardanico, line 19637: No need for .git here

packages.json
Q-Master: fixed

packages.json
Q-Master: fixed

PR #1666, opened 2020-10-12
Add package json5

sealmove:


sealmove:
Don't mind this. I realized json5 is missing a key feature so me, so I am probably not up to the task of maintaining it.

PR #1680, opened 2020-10-27
Add package kdb

inv2004:


narimiran:
```E: kdb: a package by that name already exists.```

Yardanico:
@inv2004 you added your package before already :D https://github.com/nim-lang/Nim/tree/devel/lib/std

inv2004:
I did not know - should I publish every new version or not. But now I
understand that it is not necessary - nimble grabs the code from git-head

PR #1692, opened 2020-11-04
add package oshostname

jrfondren:


jrfondren:
noteworthiness:
https://lib.rs/crates/hostname - used in 82 crates (directly), #6 in Operating Systems, 237k downloads in the last month.
https://metacpan.org/pod/Sys::Hostname - core library, used by almost every Perl script I write for sysadmin tasks.

... come to think of it, it's also arguable for inclusion into posix_utils and then os as platform supported is filled out.

PR #1726, opened 2020-11-28
Add mk

thisago:


narimiran:
@dom96 @Araq what do we think of "impostor" group/project https://gitlab.com/nimlang ?

Araq:
I think we should kindly ask @thisago to use a more descriptive project name.

thisago:
Hi @narimiran, I selected this name because it was avaliable, if you want i can change to another one. sorry

thisago:
yes, a agree, but the project is a handler to create folders and files on terminal, the name need to be short to write easier, I wrote the package name to "mk" because it is the name of command, the name as "mkfs" or "shmk" is better? sorry by my bad english. Thanks for the feedback

thisago:
Thanks for the help, guys, I'm going to change the name of the gitlab group and the package name to "shmk" and make another pull request; I recommend that you get that group name from gitlab. Thank you for your help!

PR #1728, opened 2020-11-30
Remove Inumon

dizzyliam:
I was overly optimistic about how motivated I would be to maintain this.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -18609,22 +18609,6 @@
     "web": "https://github.com/bung87/tim_sort"
   },
   {
-    "name": "inumon",
-    "url": "https://github.com/dizzyliam/inumon",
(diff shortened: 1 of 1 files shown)


narimiran:
We usually don't remove packages once added.

dizzyliam:
I can understand why you would make that call overall. However, I'm pretty sure Inumon has no users who will have their projects suddenly bricked, seeing as there have since come to be much better alternatives that are actually maintained.

Clyybber:
Maybe we should add an `unmaintained` or `deprecated` in such cases?

dizzyliam:
Actually I have just noticed the mention of the `abandoned` tag in the README. Would it be distinct from those tags?

PR #1744, opened 2020-12-09
GitHub actions

iffy:
If you want to switch to GitHub Actions for running checks, this seems to work (and doesn't hit the GitHub rate limits): https://github.com/iffy/packages/runs/1525986310?check_suite_focus=true


dom96 (approved):
Beautiful, thanks!

narimiran:
@dom96 Can we now remove Travis? It is painfully slow (it seems way slower than it was before).

PR #1752, opened 2020-12-17
Add package bisect

berquist:


narimiran:
Please add `README.md` which explains what this package is, what it does, and how one can use it.

PR #1780, opened 2021-01-11
Renamed midio_ui and midio_ui_canvas packages

kristianhasselknippe:


packages.json
```diff
@@ -20397,7 +20397,11 @@
   },
   {
     "name": "midio_ui",
-    "url": "https://github.com/nortero-code/midio-ui",
+    "alias": "denim_ui",
```
narimiran, line 20400:
```suggestion
    "alias": "denim_ui"
```

narimiran, line 20400:
```suggestion
    "alias": "denim_ui"
```

PR #1793, opened 2021-01-24
Add package nobject

Carpall:


narimiran:
A readme, with a short explanation what does this package do and how to use it, would be nice.

PR #1820, opened 2021-02-25
onnxruntime c api for nim

YesDrX:
https://github.com/YesDrX/onnxruntime-nim

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -1,6 +1,6 @@
 [
   {
-    "name": "onnxruntime_nim",
+    "name": "onnxruntime",
     "url": "https://github.com/YesDrX/onnxruntime-nim.git",
     "method": "git",
     "tags": [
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -1,4 +1,15 @@
 [
+  {
+    "name": "onnxruntime-nim",
```
dom96, line 3: Names cannot contain dashes, please remove the `-nim`.

dom96 (changes requested):

packages.json
YesDrX: just removed the dash in name

packages.json
```diff
@@ -1,4 +1,14 @@
 [
+  {
+    "name": "onnxruntime_nim",
```
dom96, line 3: The name needs to match your .nimble file's name :)

dom96 (changes requested):

packages.json
Yardanico: Isn't using "nim" in package names disallowed anyway?

dom96: nope

PR #1829, opened 2021-03-06
Add package nim_seq

Unaimend:


SolitudeSF:
why not dna_seq? this is name is half redundant, half misleading.

Unaimend:
So whats wrong with the travis status report?

Unaimend:
It's not my package so I have no idea, but I would agree that its a perfect name, but dna_seq also would not fit since we plan to implement a lot more bioinformatics related file formats which do not have anything todo with dna.
But anyways there was a way to define a different name for the nimble package, and not use the git one, wasnt there?

SolitudeSF:
> But anyways there was a way to define a different name for the nimble package, and not use the git one, wasnt there?

the name of the package is the name of nimble file.

Unaimend:
I will make the changes and make a new pull request

PR #1839, opened 2021-03-13
Add package cirru_edn

jiyinyiyong:


packages.json
```diff
@@ -14935,6 +14935,18 @@
     "license": "MIT",
     "web": "https://github.com/Cirru/writer.nim"
   },
+  {
+    "name": "cirru_edn",
+    "url": "https://github.com/Cirru/cirru-edn.nim",
```
xflywind, line 14940: it can't have dash line.

Yardanico: @xflywind that's the URL, it can

xflywind: OK, thanks

PR #1842, opened 2021-03-15
Add package nesper

elcritch:


packages.json
```diff
@@ -21862,20 +21862,36 @@
...
+    "name": "nesper",
+    "url": "https://github.com/elcritch/nesper",
+    "method": "git",
+    "tags": [
+      "esp32",
+      "esp-idf",
+      "mcu",
+      "microcontroller",
+      "embedded",
+      ""
```
narimiran, line 21891: Why an empty tag?

elcritch: Odd, I must’ve left a space at the end of the nimble tags prompt. Looks like it got turned into an empty tag.

narimiran: Ok, I removed it and solved merge conflicts. It should be ok now.

PR #1851, opened 2021-03-29
Add package nim_openapi

LantosHub:


SolitudeSF:
why is this called openapi?

LantosHub:
> why is this called openapi?

Typo, will re-upload

PR #1856, opened 2021-04-03
Add package static_server

bung87:


Yardanico:
That name is already occupied - see https://github.com/nim-lang/packages/pull/723

Yardanico:
out of curiosity - what does the fork do compared to the original that you decided to create a new package?

bung87:
it depends on my finder package https://github.com/bung87/finder, which can serve static file from memory and zip not only  file system.

PR #1918, opened 2021-05-30
Add ass.

0kalekale:
bindings for libass


narimiran:
> `E: ass has no tags`

Please add (or: do not remove) and populate the `tags` field.

PR #1936, opened 2021-06-18
Add package listenbrainz

tandy-1000:


dom96 (approved):
Nice! This is the same library that Spotify uses isn't it?

tandy-1000:
No, it is an alternative to Last.fm made by the MusicBrainz crew :)

PR #1942, opened 2021-06-23
add nim-ws package

jangko:


packages.json
```diff
@@ -23001,5 +23001,19 @@
     "description": "NWSync Repository Management utilities",
     "license": "MIT",
     "web": "https://github.com/Beamdog/nwsync"
+  },
+  {
+    "name": "nim-ws",
```
dom96, line 23006: This is incorrect, it should be `ws` but `nim-ws` already exists: https://github.com/treeform/ws.

You'll need to pick a different name, I'll revert this.

narimiran: Why this forbidden prefix `nim-` isn't checked by our CIs?

I always forget to check this manually.

dom96: probably because there are some old packages that have hyphens in the name, it's time we just get rid of these and add the check to package_scanner

jangko:
we are still hunting new name for nim-ws. see status-im/nim-ws#74

PR #1954, opened 2021-06-28
Added 'pl0t' package

al6x:


packages.json
```diff
@@ -23156,5 +23156,22 @@
...
+      "plot",
+      "chart",
+      "table",
+      "excel",
+      "spreadsheet",
+      "visualization",
+      "data"
+    ],
+    "description": "Plot and visualize data",
+    "license": "MIT",
```
dom96, line 23174: This doesn't look like MIT: https://github.com/al6x/pl0t/blob/main/license

packages.json
```diff
@@ -23156,5 +23156,22 @@
     "description": "A Nim client for Todoist's REST API",
     "license": "Apache-2.0",
     "web": "https://ruivieira.github.io/nim-todoist/index.html"
+  },
+  {
+    "name": "pl0t",
+    "url": "https://github.com/al6x/pl0t?subdir=api/nim",
```
dom96, line 23162: since this is a pl0t repo already, why the subdir?

al6x: It's monorepo, there are couple of different things, the language APIs reside in `/api` folder, so there going to be `/api/nim` and `/api/typescript`.

packages.json
al6x: Thanks, updated.

al6x:
Conflict resolved, please review.

Vindaar:
@dom96 We allow packages with a license "Proprietary" in `nimble`?

dom96:
yes, why wouldn't we?

PR #1959, opened 2021-06-29
add package mailcow

Vaipex:


dom96 (approved):

narimiran:
A `README` would be nice to have, to see what this package is, how to use it, etc.

Vaipex:
> A `README` would be nice to have, to see what this package is, how to use it, etc.

yeah, I will add some example codes and a readme in the future!

PR #1973, opened 2021-07-19
Update isocodes tags and description.

kraptor:
I added support for ISO 15924 (language scripts) and ISO 4217 (currencies) to `isocodes`.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -23332,12 +23332,18 @@
       "iso",
       "countries",
       "country",
+      "language",
+      "languages",
+      "currency",
+      "currencies",
       "3166",
       "3166-1",
       "3166-2",
-      "3166-3"
+      "3166-3",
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -23332,12 +23332,18 @@
...
+      "currency",
+      "currencies",
       "3166",
       "3166-1",
       "3166-2",
-      "3166-3"
+      "3166-3",
+      "15924",
+      "4217"
     ],
```
dom96, line 23345: You should be more explicit: `ISO-3166`, etc. for these tags.

dom96 (changes requested):

packages.json
kraptor: Updated with more explicit tags as per your request :)

PR #1977, opened 2021-07-26
Add package gnu

Knaque:


narimiran:
@dom96 is stuff like `gnu` some kind of a "reserved name" or is it up for grabs? (I have a similar concern for `cpython`: https://github.com/nim-lang/packages/pull/1976)

Knaque:
> @dom96 is stuff like `gnu` some kind of a "reserved name" or is it up for grabs? (I have a similar concern for `cpython`: https://github.com/nim-lang/packages/pull/1976)

You know, I did wonder that, but forgot to ask before publishing.

PR #1982, opened 2021-08-09
Added notcurses

michaelsbradleyjr:


packages.json
```diff
@@ -23487,5 +23487,22 @@
...
     "license": "MIT",
     "web": "https://github.com/thisago/ytextractor"
+  },
+  {
+    "name": "notcurses",
+    "url": "https://github.com/michaelsbradleyjr/nim-notcurses",
+    "method": "git",
+    "tags": [
+      "TUI",
+      "character graphics",
```
dom96, line 23497: no spaces in tags plz

dom96 (approved):
rest looks great

packages.json
michaelsbradleyjr: Fixe that, and I shortened the description since shorter descriptions seem to be favored.

michaelsbradleyjr:
Rebased.

michaelsbradleyjr:
I also added some tags: I've been working trough the [Notcurses book](https://nick-black.com/htp-notcurses.pdf) and the library's [demo code](https://github.com/dankamongmen/notcurses/tree/master/src/demo) and there are terms (and variants of them) that come up frequently.

packages.json
```diff
@@ -23516,5 +23516,57 @@
...
+      "pictures",
+      "terminal",
+      "terminals",
+      "tui",
+      "tuis",
+      "unicode",
+      "video",
+      "videos",
+      "widget",
+      "widgets"
```
dom96, line 23566: Okay, that's overdoing it... when you're adding plurals for each tag then you're going too far. You don't need this many, just:

* `tui`
* `library`
* `cli`

Is honestly enough.

dom96 (changes requested):

packages.json
michaelsbradleyjr: Alright, I'm happy to comply.

But, I wasn't on a wild tear; as I mentioned:

>  I've been working trough Notcurses' book/docs... there are terms (and variants of them) that come up frequently.

I tried out searches for various terms plural and singular, etc. on [nimble.directory](https://nimble.directory/) (not necessarily related to Notcurses) and found uneven results, so decided to expand on a base set of terms that I think are relevant to Notcurses, and so nim-notcurses:
```json
"tags": [
    "animate",
    "character",
    "cli",
    "color",
    "command-line",
    "console",
    "curses",
    "emoji",
    "glyph",
    "grapheme",
    "graphics",
    "image",
    "interface",
    "multimedia",
    "ncurses",
    "picture",
    "terminal",
    "tui",
    "unicode",
    "video",
    "widget"
 ]
```

If that's still too much, I'm happy to go with your suggestion of:
```json
"tags": [
    "cli",
    "library",
    "tui"
]
```

I was simply attempting to make nim-notcurses more find-able from [nimble.directory](https://nimble.directory/), given Notcurses touches on a lot of TUI capabilities; and because it's just package metadata, I didn't imagine it would be a problem.

I take no offense, and will modify my PR re: what you think is most suited re: nim-notcurses' `"tags": [...],"`.

dom96: Personally I don't think we should be using these tags as a way to make things more searchable, the tags should be for making things more groupable. At the end of the day nimble.directory should maybe include keywords in package descriptions in its search, we shouldn't be adding many tens of tags to fix that for it.

michaelsbradleyjr: Fair enough, then I'll change it now, as you suggested, to:
```
"tags": [
    "cli",
    "library",
    "tui"
]
```

michaelsbradleyjr: Done.

dom96 (approved):

PR #1989, opened 2021-08-21
Add package composition

DavidMeagher1:


packages.json
```diff
@@ -23529,5 +23529,16 @@
     "description": "A low-level Nim wrapper for Notcurses: blingful TUIs and character graphics",
     "license": "Apache-2.0",
     "web": "https://github.com/michaelsbradleyjr/nim-notcurses"
+  },
+  {
+    "name": "composition",
+    "url": "https://github.com/DavidMeagher1/composition",
+    "method": "git",
+    "tags": [
+      "library"
```
xflywind, line 23538: IMO you can use some more verbose tags: "Composition", ...

PR #1990, opened 2021-08-22
Added package oolib

Glasses-Neo:


packages.json
```diff
@@ -23540,5 +23540,19 @@
...
+  },
+  {
+    "name": "oolib",
+    "url": "https://github.com/Glasses-Neo/OOlib",
+    "method": "git",
+    "tags": [
+      "nim",
+      "oop",
+      "metaprogramming",
+      "nimble"
```
xflywind, line 23552:
```suggestion
      "oop",
      "metaprogramming",
```
@Glasses-Neo, maybe remove "nim" and "nimble" tags?

Glasses-Neo: That's right, thanks!

PR #1994, opened 2021-08-29
Add package watch_for_files

hamidb80:


hamidb80:
btw if you look at the commit, in `tags` section,  you'll see am empty string `""`
if think it considered a bug

packages.json
```diff
@@ -23622,5 +23622,20 @@
...
+  },
+  {
+    "name": "watch_for_files",
+    "url": "https://github.com/hamidb80/watch_for_files",
+    "method": "git",
+    "tags": [
+      "file-watcher",
+      "file",
+      "watcher",
+      "cross-platform",
```
narimiran, line 23634:
```suggestion
      "cross-platform"
```

packages.json
```diff
@@ -23622,5 +23622,20 @@
...
+      ""
```
narimiran, line 23635:
```suggestion
```

PR #1995, opened 2021-08-31
Add package nodesnim

Ethosa:


narimiran:
> `E: nodesnim: a package by that name already exists.`

You need to pick another name for your package.

xflywind:
It is already be added to packages+

https://github.com/nim-lang/packages/blob/80087047926e084dee35dc097c495be3c5827e8f/packages.json#L18795

PR #2005, opened 2021-09-10
Add package tic80

thisago:


thisago:
I used `nimble publish` before create repo... The nimble needs to inform that repo doesn't exists and cancel.

I will create a issue in `nimble` repo.

PR #2009, opened 2021-09-18
Add package surfing

momeemt:


packages.json
```diff
@@ -23740,5 +23740,19 @@
...
+    "name": "surfing",
+    "url": "https://github.com/momeemt/surfing",
+    "method": "git",
+    "tags": [
+      "base64",
+      "cli",
+      "string",
+      "surfing"
+    ],
+    "description": "A new awesome nimble package",
```
Yardanico, line 23754: Maybe you can change this to something more meaningful rather than the default description? :)

momeemt: Thanks! I fixed.

dom96 (approved):
Nice!

PR #2013, opened 2021-09-28
Add package selenimum

myamyu:


dom96 (approved):
Looks good. Note that we have at least two other packages that do the same thing already :)

myamyu:
@dom96 Thank you for review. 

> Note that we have at least two other packages that do the same thing already :)

Yes, I know them.

PR #2014, opened 2021-09-29
add description to zfdbms

pietroppeter:
I am not the author of zfdbms and taking description from About section of GitHub: https://github.com/zendbit/nim.zfdbms
This PR is to fix a broken nimble search and nimble list: https://github.com/nim-lang/nimble/issues/947

cc @zendbit

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -23800,7 +23800,7 @@
       "sqlite",
       "postgre"
     ],
-    "description": "",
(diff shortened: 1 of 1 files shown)


dom96 (approved):
Thanks for fixing this! Our CI should really catch this I suppose.

pietroppeter:
well, it seems rather easy to do a fix so that CI catches this, I will prepare a PR

PR #2015, opened 2021-09-29
add check for empty description

pietroppeter:
see #2014


pietroppeter:
CI correctly catches an error if an empty description is provided: see CI failure example https://github.com/nim-lang/packages/runs/3749330772

dom96 (approved):
<3

PR #2035, opened 2021-10-20
Add xidoc

xigoi:
The guide says that I should specify my e-mail address; where do I put it?


xigoi:
> > The guide says that I should specify my e-mail address; where do I put it?
> 
> Which guide?

https://github.com/nim-lang/packages

>While we really appreciate your contribution, please follow the requirements: other developers will rely on your package. Non-compliant packages might be removed with no warning.
>
>* ...
>* Provide a contact email address.
>* ...

narimiran:
Hah, had no idea about that. I wouldn't worry in your place :)

PR #2053, opened 2021-11-03
Add package config

vsajip:


packages.json
```diff
@@ -24135,5 +24135,19 @@
     "description": "Easily use Animate.css classes",
     "license": "MIT",
     "web": "https://github.com/thisago/animatecss"
+  },
+  {
+    "name": "config",
+    "url": "ssh://git@github.com/vsajip/nim-cfg-lib",
```
narimiran, line 24141: @dom96 are these `ssh` urls ok?

vsajip: Ah, no, sorry. Let me update to the correct https:// URLs.

PR #2055, opened 2021-11-04
Add package odsreader

dariolah:


narimiran:
A `readme` would have been nice....

Yardanico:
@dariolah please add the "zip" package to your requires in the nimble file as you're using it in your code. Also as miran said it would be a good idea to have at least a small readme :)

PR #2063, opened 2021-11-11
Add package nimkalc

nocturn9x:


narimiran:
You have already submitted that package. (Try `nimble search nimkalc` to see it)

PR #2092, opened 2021-12-21
Add package honeycomb

KatrinaKitten:


KatrinaKitten:
Apologies, I was testing the `publish` command and didn't realize it would auto-PR. I'll resubmit when the library is actually ready.

PR #2103, opened 2021-12-29
Rename socks -> socks5

FedericoCeratto:


dom96 (commented):
Do we want to set up an alias for this? is it a package that has been around for a while? https://github.com/nim-lang/packages#renaming-packages

FedericoCeratto:
Yes please.

PR #2126, opened 2022-01-22
Change username to resolve name problem

EchoPouet:
Problem with the current user name to close to my current job.

No problem to other projects because nobody use my work :)

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -22061,7 +22061,7 @@
   },
   {
     "name": "checkpack",
-    "url": "https://gitlab.com/ArMour85/checkpack",
+    "url": "https://gitlab.com/EchoPouet/checkpack",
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -24871,5 +24871,21 @@
...
+      "clipboard",
+      "remote",
+      "server",
+      "utility",
+      "cli",
+      "tool"
+    ],
+    "description": "Manage your clipboard remotely",
+    "license": "MIT",
+    "web": "https://gitlab.com/lurlo/servclip"
```
dom96, line 24889: Unintended change?

EchoPouet: Sorry, bad rebase, I'll fix it.

PR #2163, opened 2022-02-27
added nofi package

ct-clmsn:
would like to add the [nofi package](https://github.com/ct-clmsn/nofi) to the nimble manifest.


ct-clmsn:
made [PR](https://github.com/nim-lang/packages/pull/2164) that adds string support for boost

PR #2166, opened 2022-03-07
Add package iterrr

hamidb80:


packages.json
```diff
@@ -25371,5 +25371,21 @@
     "description": "Tool to download dogs images",
     "license": "Apache-2.0",
     "web": "https://github.com/thexxiv/dogapi-cli"
+  },
+  {
+    "name": "iterrr",
+    "url": "https://github.com/hamidb80/steps",
```
xflywind, line 25377: Hello, @hamidb80, It seems to be https://github.com/hamidb80/iterrr now.

PR #2178, opened 2022-03-21
Add package aqcalc

VitorGoatman:


xflywind:
The package has been already added: https://github.com/nim-lang/packages/pull/2176

If you need to change the descriptions of your package, please edit it directly.

PR #2205, opened 2022-04-19
Add govee

nonimportant:
https://github.com/nonimportant/nim-govee


dom96:
oh cool, I own a Govee LED strip :D

PR #2216, opened 2022-04-28
Update packages.json - add periapsisEngine

KnedlikMCPE:


packages.json
```diff
@@ -1,4 +1,13 @@
 [
+  {
+    "name": "periapsisEngine",
+    "url": "https://github.com/Periapsis-Studios/Periapsis-Engine",
+    "method": "git",
+    "author": "Knedlik",
```
xflywind, line 6: You have to specifiy tags attributes, otherwise => "periapsisEngine has no tags"

KnedlikMCPE: Should hopefully be okay now.

PR #2242, opened 2022-05-24
add package mouse

hiikion:


narimiran:
> `E: mouse has an unexpected license: "MPL2"`

PR #2243, opened 2022-05-27
Added nim_tinyfiledialogs package

Patitotective:
:]

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -4,9 +4,8 @@
     "url": "https://github.com/Patitotective/nim_tinyfiledialogs", 
     "method": "git", 
     "tags": [
-      "wrapper", 
-      "bindings", 
-      "tinyfd"
+      "dialogs", 
+      "file dialogs"
     ], 
     "description": "Tiny file dialogs Nim bindings.", 
     "license": "MIT", 
--- a/packages.json
+++ b/packages.json
@@ -1,7 +1,7 @@
 [
   {
-    "name": "nim_tinyfiledialogs", 
-    "url": "https://github.com/Patitotective/nim_tinyfiledialogs", 
+    "name": "tinyfiledialogs", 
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -1,4 +1,17 @@
 [
+  {
+    "name": "nim_tinyfiledialogs", 
+    "url": "https://github.com/Patitotective/nim_tinyfiledialogs", 
+    "method": "git", 
+    "tags": [
+      "wrapper", 
+      "bindings", 
+      "tinyfd"
```
Yardanico, line 9: i think instead of generic tags like this you can add tags like:
`file dialogs`
`dialogs`


 so e.g. on nimble.directory it'll be much easier to find your package

dom96:
I would suggest renaming this to `tinyfiledialogs`, the fact it's a Nim package is already clear by the fact that it's in Nimble and `import nim_tinyfiledialogs` is too verbose :)

packages.json
```diff
@@ -1,4 +1,16 @@
 [
+  {
+    "name": "tinyfiledialogs", 
+    "url": "https://github.com/Patitotective/tinyfiledialogs.nim", 
+    "method": "git", 
+    "tags": [
+      "dialogs", 
+      "file dialogs"
```
dom96, line 8: Tags should be one word.

```suggestion
      "file-dialogs"
```

dom96 (approved):

xflywind:
> E: tinyfiledialogs: a package by that name already exists.

Patitotective:
Yea, I just found a package that wraps the same library :[
Yet, my bindings do include a color picker so I think I'll just rename them to tinyfiledialogs.nim (?) (~~verbose~~)

Patitotective:
@dom96 Should I open another PR to fix the name conflict or?

dom96:
this is what I get for not waiting for CI lol, sorry

PR #2264, opened 2022-06-28
Add package seq2d

avahe-kellenberger:


narimiran:
> E: seq2d has an unexpected license: "GPL-2.0-only"

avahe-kellenberger:
Why is that unexpected?

avahe-kellenberger:
As far as I'm aware, it's the proper identifier from SPDX https://spdx.org/licenses/GPL-2.0-only.html

But I don't know what is expected by nimble, here

PR #2271, opened 2022-07-08
add library wttrin

Infinitybeond1:
i recently created a library with a few functions scraping weather data from wttr.in, this is very useful because it does not require an open weather map api key

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -26656,4 +26656,19 @@
     "license": "MIT",
     "web": "https://github.com/bob16795/hangover"
   }
+  {
+    "name": "wttrin",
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -26655,5 +26655,20 @@
...
+  },
+  {
+    "name": "wttrin",
+    "url": "https://github.com/Infinitybeond1/wttrin",
+    "method": "git",
+    "tags": [
+      "weather",
+      "weather-api",
+      "cli",
+      "nim",
```
xflywind, line 26667:
```suggestion
```

PR #2297, opened 2022-07-28
Add package util

thisago:


packages.json
```diff
@@ -26886,5 +26886,20 @@
...
+  },
+  {
+    "name": "util",
+    "url": "https://github.com/thisago/util",
+    "method": "git",
+    "tags": [
+      "html",
+      "utility",
+      "nim",
+      "nim-lang",
```
ringabout, line 26898:
```suggestion
```

Hi, the two tags don't seeem to be necessary.

thisago: oh, this is because I copied project tags in github, i will fix

PR #2305, opened 2022-08-03
Add package yamlscript

ingydotnet:


ringabout:
@ingydotnet Hello, https://github.com/yaml/yamlscript/tree/main/nim doesn't have any code in it. Did you publish it by accident? You should also specify the subdir like https://github.com/yaml/yamlscript?subdir=nim.

ingydotnet:
Thanks for the subdir tip. Was wondering. I'll add some code. Cheers.

PR #2313, opened 2022-08-05
Add package geomancer

VitorGoatman:


ringabout:
Hi @VitorGoatman , https://github.com/nim-lang/packages/pull/2312 has been merged. You don't need to use `nimble publish` anymore and instead tag your package properly.

PR #2317, opened 2022-08-09
Rename compiler to nim

yyoncho:
- see https://github.com/nim-lang/nimble/pull/1017 and https://github.com/nim-lang/Nim/pull/20179

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -2938,15 +2938,20 @@
     "web": "https://github.com/NimParsers/parsetoml"
   },
   {
-    "name": "compiler",
+    "name": "nim",
     "url": "https://github.com/nim-lang/Nim.git",
     "method": "git",
     "tags": [
-      "library"
+      "library",
+      "compiler"
     ],
-    "description": "Compiler package providing the compiler sources as a library.",
+    "description": "Compiler package providing the compiler sources as a library/binary.",
     "license": "MIT",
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -2938,16 +2938,21 @@
...
+      "compiler"
     ],
-    "description": "Compiler package providing the compiler sources as a library.",
+    "description": "Compiler package providing the compiler sources as a library/binary.",
     "license": "MIT",
     "web": "https://github.com/nim-lang/Nim"
   },
+  {
+    "name": "nim",
+    "alias": "compiler"
```
dom96, line 2954: This needs to be the other way around

yyoncho: Fixed.

packages.json
```diff
@@ -2938,16 +2938,21 @@
...
+    "name": "nim",
     "url": "https://github.com/nim-lang/Nim.git",
     "method": "git",
     "tags": [
-      "library"
+      "library",
+      "compiler"
     ],
-    "description": "Compiler package providing the compiler sources as a library.",
+    "description": "Compiler package providing the compiler sources as a library/binary.",
```
zah, line 2948:
```suggestion
    "description": "Package providing the Nim compiler binaries plus all its source files that can be used as a library",
```

zah (approved):

PR #2323, opened 2022-08-13
added zigcc

enthus1ast:


narimiran:
@dom96 Should a single command (`discard execShellCmd "zig cc " & commandLineParams().join(" ")`) be a nimble package?

dom96:
I don't see why not. In this case it makes using `zig cc` with Nim much easier.

PR #2324, opened 2022-08-15
Add package swayipc

nnsee:


nnsee:
wow, i wish i knew this package existed before writing one from scratch

PR #2327, opened 2022-08-18
Add clipper2

scemino:


dom96 (commented):
why not just name it "clipper2"?

scemino:
> why not just name it "clipper2"?

I didn't know if I could use this name, I'm not the author of the official C++ library, but if it's ok I can change the name.

dom96:
I'm fairly sure it's fine to use the same name for something that is wrapping that library.

PR #2331, opened 2022-08-24
Add metacall.

viferga:


narimiran:
Is this a Nim package? (cc @dom96)

viferga:
> Is this a Nim package? (cc @dom96)

It is a nim wrapper to a c library. A port of that library.

PR #2345, opened 2022-09-02
Add package powernim

wreedb:


narimiran:
> E: powernim has an unexpected license: "BSD-2-Clause"

cc @dom96

PR #2350, opened 2022-09-08
Add package bitfields

adokitkat:


Yardanico:
Hey, just wanted to ask out of curiosity (not related to the actual PR) - do you know about https://nim-lang.org/docs/bitops.html and how does your module differ from it?

adokitkat:
> Hey, just wanted to ask out of curiosity (not related to the actual PR) - do you know about https://nim-lang.org/docs/bitops.html and how does your module differ from it?

Hello @Yardanico, yes I do know about std/bitops. Bitops works directly with integers and maybe I'll even use it later in my package but the difference is bitfields convert integers to seq of Bits ( = range[0'u8..1'u8]). This way I can create arbitrary long sequences of bits and use std/sequtils on them too.

PR #2352, opened 2022-09-08
Add package bitseqs

adokitkat:


adokitkat:
@narimiran sorry to disrupt you but is it possible to change url and web to "https://github.com/adokitkat/bitfields" ? I renamed the repository but it seems like it didn't change when doing nimble publish.

narimiran:
Send another PR, fixing that.

narimiran:
But, on further inspection, I don't understand what you're saying.

> is it possible to change url and web to "https://github.com/adokitkat/bitfields" ?

It is already that link!?

PR #2362, opened 2022-09-19
Adds dimage

d-Pixie:
Adds dimage, a small library that I extracted from a larger project. It allows you to pull mime type and size information from images. It is written in pure Nim and has cligen as its only dependency except for the Nim stdlib.


narimiran:
Ci reports this:

> `E: dimage has an unexpected license: "LGPL-3.0-only"`


Changing it to `"lgpl-3.0"` should do the trick.

Issue #2300, opened 2022-07-29
Empty tags


Patitotective:
I'm working on an application that uses the packages list (`packages.json`) and I've found an issue (i think).
The following packages have an empty tag:
```json
{"name":"htsparse","url":"https://github.com/haxscramper/htsparse","method":"git","tags":["library","wrapper","parser",""],...}
{"name":"syllables","url":"https://github.com/tonogram/nim-syllables","method":"git","tags":["library","language","syllable","syllables",""],...}
{"name":"drawim","url":"https://github.com/GabrielLasso/drawim","method":"git","tags":["draw","drawing","gamedev",""],...}
{"name":"scinim","url":"https://github.com/SciNim/scinim","method":"git","tags":["scinim",""],...}
{"name":"tmpnim","url":"https://github.com/ment1na/tmpnim","method":"git","tags":["library","tmpfs","ramdisk","tempfile","linux",""],...}
```
This isn't a critic issue but I think `package_scanner.nim` should check that.

ringabout:
fixed by https://github.com/nim-lang/packages/pull/2385

ringabout:
fixed by https://github.com/nim-lang/packages/pull/2390

PR #2385, opened 2022-10-14
Check for empty tags

Patitotective:
As I said in #2300, some packages have empty tags, this PR modifies `packages_scanner.nim` so that it warns when packages have empty tags.

Since I'm already making this PR I wanted to know if I should add the following changes to this PR or make a new: add a `config.nims` file with `--define:ssl` and check for valid `JsonNode` kinds (check that `tags` is a `JArray` of `JString`s, `url` is a `JString` and so on).

:heart:

### DIFF
--- a/package_scanner.nim
+++ b/package_scanner.nim
@@ -7,6 +7,7 @@
 #  * Missing/unknown method
 #  * Missing/unreachable repository
 #  * Missing tags
+#  * Empty tags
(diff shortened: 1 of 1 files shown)


ringabout:
Hello, @Patitotective  Following prs that removes the empty tags and turn the warning into an error is appreciated!

PR #2390, opened 2022-10-19
Empty tags now count as errors + removed empty tags from the packages list

Patitotective:
Related to #2385, #2300.


ringabout (approved):
Nice!

PR #2391, opened 2022-10-19
-d:dontFetchRepos and config.nims

Patitotective:
- Created `config.nims` with `--define:ssl` for easier compilation.
- Added a `-d:dontFetchRepos` (devel only) flag that makes `package_scanner.nim` don't call `canFetchNimbleRepository`. This is useful for testing changes without having to fetch over 2000 repositories (especially with slow internet).

### DIFF
--- a/config.nims
+++ b/config.nims
@@ -1,0 +1,1 @@
+--define:ssl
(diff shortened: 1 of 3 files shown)


narimiran:
Pinging @dom96 once again.

dom96:
Feel free to merge these kinds of PRs

PR #2394, opened 2022-10-27
Add package Pigeon

dizzyliam:


dom96 (approved):
Awesome package.

Btw consider putting your `request` module inside a `pigeon` sub-folder so that others cannot import it via `import request` (i.e. it would clash with another package called `request`).

PR #2411, opened 2022-11-29
Add package template

ArikRahman:
Basic nim template for getting started working on nim projects in codium

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -27944,5 +27944,16 @@
     "description": "Helper procs for CGI applications in Nim.",
     "license": "MIT",
     "web": "https://github.com/nim-lang/cgi"
+  },
+  {
+    "name": "template",
+    "url": "https://github.com/ArikRahman/template",
+    "method": "git",
+    "tags": [
+      "y"
+    ],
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -27944,5 +27944,16 @@
     "description": "Helper procs for CGI applications in Nim.",
     "license": "MIT",
     "web": "https://github.com/nim-lang/cgi"
+  },
+  {
+    "name": "template",
+    "url": "https://github.com/ArikRahman/template",
+    "method": "git",
+    "tags": [
+      "y"
```
narimiran, line 27953: Can we get something more meaningful than this?

Yardanico:
I assume you want to create a template of a Nim project. If so, I don't think publishing it as a Nimble package is a good idea unless your Nimble package is an application that _creates_ that structure. If you just want to make a template, make it a repository and spread around, also maybe by using GitHub template repo functionality. For example see https://github.com/treeform/nimtemplate

ArikRahman:
I'll try working on the former, sounds intriguing, thanks.

PR #2446, opened 2022-12-31
rename nimp5 to p5nim + alias + change repo

pietroppeter:
I picked up the old nimp5 repo from @Foldover and started fixing and adding stuff. Also it was renamed to p5nim (from nimp5)
He is ok to remove his version and add mine: https://github.com/Foldover/nim-p5/pull/2#issuecomment-1336357474

https://github.com/pietroppeter/p5nim

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -9875,8 +9875,8 @@
     "web": "https://github.com/numforge/nim-clblast"
   },
   {
-    "alias": "nimp5",
-    "name": "p5nim"
+    "name": "nimp5",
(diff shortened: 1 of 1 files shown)


pietroppeter:
cannot tell if the failure is due to an issue in this PR or if it is unrelated...

pietroppeter:
Ah now I see, the error is indeed in the PR I made a mistake in the alias fields (name and alias should be exchanged). Will fix

ringabout:
Nice!

PR #2447, opened 2023-01-01
update user name

ShalokShalom:
this broke previously, since the user changed its account


Yepoleb:
Looks like the first part of your PR has already been applied. Can you rebase this to only include the repo rename?

ShalokShalom:
thanks 👍🏻

PR #2448, opened 2023-01-02
Include pape lib

hdbg:


packages.json
```diff
@@ -26508,6 +26508,9 @@
     "name": "capstone",
     "url": "https://github.com/hdbg/capstone-nim",
     "method": "git",
+    "tags": ["wrapper", "disasm"],
+    "description":"Capstone3 high-level wrapper",
+    "license": "MIT",
     "tags": [
       "wrapper",
       "disasm"
```
ringabout, line 26516: Looks duplicated?

hdbg: Dunno, vs code showed an error so I just fixed it

PR #2471, opened 2023-01-27
Add package scfg

xoich:


Yepoleb:
Workflow fails because of license "CC-BY-SA 4.0"

PR #2480, opened 2023-02-02
Add package colortest

z-kk:


Yepoleb:
We considered this too small of a package to be added. It seems more like a personal test project than a module useful to a broader audience.

PR #2482, opened 2023-02-02
Add spinners package

thechampagne:


Yardanico:
Hi, your package is fine, but I just wanted to comment that there's actually a similar native-Nim package (that doesn't need another library) - https://github.com/Yardanico/spinny, which is a fork of https://github.com/molnarmark/spinny that's a bit more updated.

thechampagne:
I'm aware of that 👍.

PR #2506, opened 2023-02-20
Add package nimtranslate2

777shuang:


Yepoleb:
It's unclear in what way this is different to the original package. Please clarify this in your readme and ideally work with the original author to merge back improvements.

PR #2507, opened 2023-02-21
Add package bibleTools

thisago:


thisago:
Shalom, this is a long history haha, I needed to make this lib because I made a online bible (https://git.ozzuu.com/ozzuu/bible) that uses MyBible modules, but the modules can have any identification for books and for comparing verses is extremely painful. Some books have invalid chars on it, like "Esth Gr" or "1 Sm". The lib removes anythink that isn't a letter or number, but for testing, I need to use some possibilities to check if works. I need to move this to a test file.

thisago:
Update: Now I moved the test to a test file.

PR #2515, opened 2023-02-28
Add licenses bsd-2-clause and cc-by-sa 4.0 to package scanner

Yepoleb:
Fixes checks for #2345 and #2471


ringabout:
The licence name is a bit tricky.

PR #2520, opened 2023-03-01
Improve url checking in package scanner

Yepoleb:
Use "url" field instead of "web" for checking for nimble file.
Ignore packages marked as abandoned.
Always check url if it is present, even if other fields have errors.

### DIFF
--- a/package_scanner.nim
+++ b/package_scanner.nim
@@ -131,9 +131,11 @@ proc verifyAlias(pkg: JsonNode, result: var int) =
 proc check(): int =
   var name: string
   var names = initHashSet[string]()
+  let abandonedTag = newJString("abandoned")
 
   for pkg in parseJson(readFile(getCurrentDir() / "packages.json")):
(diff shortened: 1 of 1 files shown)


package_scanner.nim
```diff
@@ -181,6 +179,12 @@ proc check(): int =
           echo "E: ", name, " has ", emptyTags, " empty tags"
           inc result
 
+      if pkg.hasKey("url"):
+        let abandoned = abandonedTag in pkg["tags"].getElems
+        if not (defined(dontFetchRepos) or abandoned):
+          if not canFetchNimbleRepository(name, pkg["url"]):
+            echo "W: Failed to fetch source code repo for ", name
```
ringabout, line 186:
```suggestion
        when not defined(dontFetchRepos):
          if abandoned or not canFetchNimbleRepository(name, pkg["url"]):
            echo "W: Failed to fetch source code repo for ", name
```

Yepoleb:
I am working on a full replacement, this will not be merged anymore.

PR #2533, opened 2023-03-09
Add package asyncIters

SirNickolas:
Like `std/asyncmacro` but for iterators, not for procs. Works with `asyncdispatch`; compatibility with `chronos` is work in progress.


ringabout (approved):
Nice work!

PR #2537, opened 2023-03-14
Add package nclip

4zv4l:
A simple wrapper around the winapi to control the clipboard

https://github.com/4zv4l/nclip


ringabout:
It wold be better to add a license file.

4zv4l:
just did, thanks !

PR #2540, opened 2023-03-16
Add package nimpretty_t

tobealive:
might be an unwanted child. But still a child.


Yepoleb:
This is cursed, but we don't have a rule against that, so I have to accept this :smile:.

tobealive:
Thanks. I hope that more than just his mother can love it.

PR #2549, opened 2023-03-24
Update nimetry, algebra links

ZoomRmc:
User @ijneb renamed to @refaqtor
Importantly for https://github.com/nim-lang/packages/issues/1807:
In this case the old username is occupied so automatic tools would incorrectly show the packages as deleted, when they've been moved!

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -13085,7 +13085,7 @@
   },
   {
     "name": "nimetry",
-    "url": "https://github.com/ijneb/nimetry",
+    "url": "https://github.com/refaqtor/nimetry",
     "method": "git",
     "tags": [
       "plot",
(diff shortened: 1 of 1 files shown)


ringabout:
@ZoomRmc Hi, it seems that the "deleted" tag should be removed as well.

ZoomRmc:
> @ZoomRmc Hi, it seems that the "deleted" tag should be removed as well.
> 
> image

Oops, sorry. It's already merged, should I open another PR?

ZoomRmc:
Will do, but not to much use, nimetry depends on a broken flippy - some of the dependencies of that are not restricted enough, it seems.

PR #2555, opened 2023-03-31
Update license for `webui`

neroist:


pietroppeter:
CI failure is unrelated (due to the fact that it was running in the brief period where choosenim stable was broken)

PR #2568, opened 2023-04-15
remove old package nodem

al6x:
It's old not supported and nobody uses it anyway


Yepoleb:
We do not remove package entries entirely. Please add the tag "abandoned" if you no longer maintain the code or "deleted" if the source repository no longer exists.

al6x:
Ok

PR #2576, opened 2023-04-25
result -> results

arnetheduck:


packages.json
```diff
@@ -12847,17 +12847,21 @@
...
+  },
+  {
+    "name": "results",
+    "url": "https://github.com/arnetheduck/nim-results",
     "method": "git",
     "tags": [
       "library",
       "result",
       "errors",
       "functional"
```
Yardanico, line 12860: Just a thought - maybe also add the tag `option` so there's a higher possibility of people finding the package?

arnetheduck: 6e9e61c8fe8906c350ace4038836be22ef05ca39

ringabout:
Weird, it breaks the Nim CI, probably some hack in the testament.

arnetheduck:
> Weird, it breaks the Nim CI, probably some hack in the testament.

`results` was previously available as `import result` (singular) - it's been deprecated for years but was removed as part of the rename

PR #2595, opened 2023-05-11
Add package gamepad

konsumer:


ringabout:
The package has been added already. No need to publish packages when the version is updated, which is handled automatically as long as the tag is updated correctly.

konsumer:
Sorry, new to nim.

PR #2597, opened 2023-05-13
Add pretty

treeform:


Yardanico:
Is there a reason to have https://github.com/treeform/pretty as a separate package from https://github.com/treeform/print? Even the README seems to be the same.

treeform:
I am rewriting the library. Much of the underlaying code has changed.
I also would like to rename the lib to stay with my `y` theme. 
I will be deprecating the print library, but I don't want to break it for people that use it now.

treeform:
I have added more things to the readme, they look different now.

PR #2604, opened 2023-05-21
remove stripekit

vfehring:
Remove stripekit as it was not meant to be published yet.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -30231,17 +30231,5 @@
     "license": "AGPL-3.0-only",
     "web": "https://github.com/Gruruya/nmostr",
     "doc": "https://gruruya.github.io/nmostr"
-  },
-  {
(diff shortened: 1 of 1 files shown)


Yepoleb:
Do you want to make this public again soon? Leaving the entry in causes less trouble than re-adding it in a few days.

vfehring:
Yeah, I would be looking to re add it again, but I’m not sure how long it will be. If leaving the entry in is beneficial, then we can do that, and I’ll just squash the repo at this time so there isn’t anything there to download I guess

Yepoleb:
It's simply less maintaining effort. I'd recommend just adding a note to the Readme that says your project is not finished. Anyone who still downloads it is on their own.

vfehring:
Fair enough. We can close this then.

PR #2697, opened 2023-08-30
Add package safeseq

avahe-kellenberger:
Seq that can safely add and remove elements while iterating.

https://github.com/avahe-kellenberger/safeseq


narimiran:
It is already in the packages list.

PR #2767, opened 2023-12-07
Add package moveiterators

sls1005:
Special iterators that use move semantics

https://github.com/sls1005/moveiterators

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -32092,7 +32092,8 @@
       "whisper.cpp"
     ],
     "description": "Bindings for Whisper.cpp",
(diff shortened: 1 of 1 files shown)


packages.json
```diff
@@ -32092,7 +32092,17 @@
       "whisper.cpp"
     ],
     "description": "Bindings for Whisper.cpp",
-    "license": "MIT",
-    "url": "https://github.com/maleyva1/whisper",
```
narimiran, line 32096: I think `url` line should have not been deleted.

PR #2778, opened 2023-12-25
Add package icecream

hmbemba:
nim port of the icecream python library

https://github.com/hmbemba/icecream


narimiran:
Sorry for the long wait, I was on holidays.

PR #2779, opened 2023-12-26
Add package nimstrict

jmsapps:
A linter for enforcing case sensitivity locally without disturbing the case insensitivity of the nim language itself

https://github.com/jmsapps/nimstrict


jmsapps:
compiler already contains this with --styleCheck

PR #2899, opened 2024-06-30
Add package lsblk

madonuko:
List out block-devices, including disks, partitions and their mountpoints

https://github.com/FyraLabs/lsblk.nim


narimiran:
> `E: lsblk has an invalid method`

This needs to be `git`, not an empty string.

PR #2915, opened 2024-08-12
Add package nclap

AinTEAsports:
A simple clap-like command line argument parser written in Nim

https://github.com/AinTEAsports/nclap


ringabout:
The package has been added. See also https://github.com/nim-lang/packages?tab=readme-ov-file#releasing-a-new-package-version

PR #2976, opened 2025-01-24
added: ormin2

Luteva-ssh:
Added package ormin2, which is the ormin package for use with nim version 2.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -34541,5 +34541,23 @@
     "description": "Non-mature SIMD-accelerated drop-ins for std/strutils functions",
     "license": "MIT",
     "web": "https://github.com/xTrayambak/kaleidoscope"
+  },
+  {
+    "name": "ormin2",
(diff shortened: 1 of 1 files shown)


narimiran:
Why a separate package and not a PR to the original?

Luteva-ssh:
Just because I can't say if changes will break some code in older/legacy projects.
ormin is based on nim version >= 0.17.2 ! I don't know if the changes will work in such an old nim version...?!?

narimiran:
> ormin is based on nim version >= 0.17.2

Nim 0.17.2 was released in 2017. Nobody (in their right mind) is using it anymore.

IMO, feel free to bump the minimal Nim version to 1.6.x in the original repo. And that version should be easy to manually test if everything still works. (If not, use the `when (NimMajor, NimMinor) >= 2.0: ...` idiom)

Luteva-ssh:
Ok. Thx! I'll bump the Version to 1.6 and test the lib with that nim version.

PR #2994, opened 2025-03-10
Add DOGEN

roger-padrell:


roger-padrell:
Why do I get the error? 
I think it is correct...

packages.json
```diff
@@ -34716,5 +34716,21 @@
...
+  },
+  {
+    "name": "dogen",
+    "url": "https://github.com/roger-padrell/dogen",
+    "method": "git",
+    "tags": [
+      "dogen",
+      "documentation",
+      "dog",
+      "docgen"
```
SolitudeSF, line 34728:
```suggestion
      "docgen",
```

PR #3000, opened 2025-03-15
Add "gummienten"

roger-padrell:


narimiran:
Should this be in the official package list without having any code written?

roger-padrell:
Sorry for that, the project was just initiated.
The project is in development now, with code.

roger-padrell:
Ended project

PR #3013, opened 2025-03-31
replace chroma source

morganholly:
treeform/chroma has not had a release in 2 years, and has multiple bugs with oklab to rgb conversion which have already been fixed. my fork bumps the version so the fixes are applied by default, vs including `@#head`.

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -10825,7 +10825,7 @@
   },
   {
     "name": "chroma",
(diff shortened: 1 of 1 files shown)


morganholly:
closed while i try to talk to treeform and try to get the original repo working right

Araq:
Thank you! That's the spirit!

PR #3041, opened 2025-05-14
add jill

capocasa:
Add jill, A nimish interface to the jack audio server


packages.json
```diff
@@ -35286,5 +35286,18 @@
     "description": "Fur is a pure Nim set of finite impulse response filters (FIR) for realtime use.",
     "license": "MIT",
     "web": "https://github.com/capocasa/fur"
+  },{
```
narimiran, line 35289:
```suggestion
  },
  {
```

PR #3052, opened 2025-06-10
add rtthread

capocasa:
A one-proc library to create posix realtime scheduling threads and some instructions


capocasa:
I believe there was a syntax error present in packages.json introduced before this pull request, second commit of this request fixes it

PR #3084, opened 2025-07-26
Add nim-compose and yaclap

ShalokShalom:
This adds packages for https://codeberg.org/emanresu3


hugosenari:
@ShalokShalom you named nim-compose as yaclap, including yaclap twice

Issue #2995, opened 2025-03-11
Cross Site Scripting in Nimble Directory website


spearphishing:
```
https://nimble.directory/search?query="><script>alert(1)</script>
```

narimiran:
cc @FedericoCeratto

spearphishing:
I apologize for the lack of explanation, but I honestly couldn't think of much to say about it.

navid-m:
Doesn't seem to be maintained anymore. Use npacks.pages.dev instead.

esafak:
Federico has moved the repo to https://codeberg.org/FedericoCeratto/nim-package-directory but it has seen no development there either.

@navid-m Where is the source for your site? Let's keep it public so we don't have a repeat of the problem.

ringabout:
You can also use 

https://nimpkgs.dayl.in
https://github.com/nimpkgs

esafak:
We really need one directory that is maintained and open to submissions; we shouldn't have to hound its owner.

@daylinmorgan Are you open to patches for nimpkgs?

FedericoCeratto:
I'll try to fix the issue but if there are volunteers that want to run the service please email me.

PR #3086, opened 2025-07-30
recommendation of https://nimpkgs.dayl.in

ringabout:
ref https://github.com/nim-lang/packages/issues/2995

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,10 +1,10 @@
-# Nim packages [![Build Status](https://travis-ci.org/nim-lang/packages.svg?branch=master)](https://travis-ci.org/nim-lang/packages)
+# Nim packages
 
 This is a central listing of all packages for
 [Nimble](https://github.com/nim-lang/nimble), a package manager for the
 [Nim programming language](http://nim-lang.org).
 
-An overview of all packages is available at https://nimble.directory.
+An overview of all packages is available at https://nimble.directory or https://nimpkgs.dayl.in.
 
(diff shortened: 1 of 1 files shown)


hugosenari:
> [It's not going to be open sourced](https://github.com/nim-lang/packages/issues/2995#issuecomment-3136864855) so I disapprove.

We are recommending a URL (an actual hosted site) so it doesn't matter if it is open or closed source, unless we are actually crowdfunding its operations.

daylinmorgan:
> [It's not going to be open sourced](https://github.com/nim-lang/packages/issues/2995#issuecomment-3136864855) so I disapprove.

Over on the other issue navid-m was saying npacks.pages.dev won't be open sourced.

nimpkgs is already open-source.

esafak:
It does matter, because we are implicitly endorsing the web site we should _contribute to_. The web site does not merely dump the package.json but makes it searchable, usable. The existing directories have problems which we would like to rectify; that's why we're having this debate. Otherwise we could continue to use nimble.directory, which is still up, after all.

esafak:
I see you set up a https://github.com/nimpkgs/crawler to automatically synchronize it with https://github.com/nim-lang/packages, so users still interact only with the latter. That is fine.

> I'm open to PRs to the website (though I would expect first some discussion about issues or feature requests).

Would you like to open up Discussions in https://github.com/nimpkgs/website to hold these discussions or do you want people to use Issues for suggestions too?

PR #3095, opened 2025-08-03
add mmops

capocasa:
mmops is a convenient AVX2 library


tags
ringabout, line 1: Please remove this file

capocasa: done. sorry!

PR #3101, opened 2025-08-11
Remove https://github.com/al6x/pl0t

esafak:
The project is "unsupported", and the web site that serves and documents it is down.

Created in response to https://forum.nim-lang.org/t/13305 

cc. @al6x OK?


al6x:
Hi, yes, its abandoned.

PR #3126, opened 2025-09-22
nim-libp2p has moved

fryorcraken:


narimiran:
The `"web"` field, few lines below, probably also needs changing.

(And `libp2pdht` package right below has also moved, if you want to change that too.)

fryorcraken:
cc @emizzle for the dht one.

fryorcraken:
@narimiran. updated!

PR #3322, opened 2026-04-10
followup(#3321): remove trailing comma

daylinmorgan:


ringabout:
If it is causing problems somewhere, it would be great to have the rule written into https://github.com/nim-lang/packages/blob/master/package_scanner.nim

daylinmorgan:
Apparently the parser in the std library is tolerant of commas so integrating into package_scanner would be a challenge. We could just have the CI test if it's spec compliant json by parsing it with jq. 

Also running `nim r pretty_json.nim` removes the commas.

See: https://github.com/nim-lang/Nim/issues/16928

PR #3360, opened 2026-05-17
Add package bau

bitstormFA:
A modern build tool for Nim

https://github.com/bitstormFA/bau


bitstormFA:
Closing: bau is already registered in Nimble packages, and the v0.5.0 release is available through the existing package entry/tag.

PR #3365, opened 2026-05-24
Setup per project folder metadata and tools

elcritch:
The goal of this PR is to enable transitioning to a folder-per-package shape from which packages.json is generated. It still supports adding packages directly to packages.json (for now).

This would make adding new packages in the future easier and avoid json clobbering. For now we will need to synchronize pkgs/ <-> packages.json.

To do that, this PR introduces sharded package metadata under `pkgs/<first-letter>/<package-name>.json` while keeping the existing top-level `packages.json` workflow compatible for now. E.g. current `nimble publish` PRs should keep working.

- A new `package_index.nim` tool now handles splitting, combining, and push-time synchronization between `packages.json` and `pkgs/`.
- CI was also reworked around that model. PR validation now uses `package_scanner.nim --check-pr --check-urls` directly against the git merge base, enforcing that PRs either add new packages or modify existing ones, but not both.
- Push handling moved into `merge.yml`, which runs only on `master`, syncs whichever side changed, validates the resulting manifest, and auto-commits the generated counterpart when needed.


elcritch:
@ringabout and @Araq what do you think of this? And grr, gotta cleanup merge scraps.

elcritch:
This also adds a `./package_index add` and a `./package_index create` for new packages. The create prompts for the data. Then `./package_index remove foo` which removes a package.

elcritch:
@ringabout @Araq if we get this merged we can start going to collecting better package metadata!

hugosenari:
Is a good moment to ask about package namespace?
Ie, instead of per package, have it per owner
docker: __org__/_package_
npm: @__org__/_package_ (org was added to v2 or something)
mvn: groupId=__com.github.owner__&artifactId=_package_
nuget: __org__._package_
pypi: _package_
CPAN: _package_
cargo: _package_
PURL: pkg:nimble/__org__/_package_@_version_?args=vals#subpath (but org is optional)

PR #3374, opened 2026-06-20
Add package hate5d

Devdeczin:
häte5d: a lightweight moddable 2D game engine for Nim

https://github.com/MarkArchive/hate5d


narimiran:
It's already in the package list.

PR #3379, opened 2026-06-25
add nimbox

capocasa:
nimbox is a process-level filesystem sandbox backed by OS-native primitives: Linux Landlock, macOS Seatbelt, and Windows ACLs.

- Single Nim API across all three platforms: `restrict(writable, read)` and `forkNimbox`/`exec`/`wait`
- Similarly-named CLI: `procbox restrict RWPATH... [--ro ROPATH...] -- CMD ARGS...`
- Repo: https://github.com/capocasa/procbox
- MIT licensed

### DIFF
--- a/packages.json
+++ b/packages.json
@@ -40523,5 +40523,19 @@
     "description": "Ergonomic database connectors for Nim.",
     "license": "BSD-3-Clause",
     "web": "https://docs.penguinite.dev/exa/"
+  },
+  {
+    "name": "sandwall",
+    "url": "https://github.com/capocasa/sandwall",
+    "method": "git",
+    "tags": [
+      "landlock",
+      "seatbelt",
+      "sandbox",
+      "security"
+    ],
+    "description": "A process-level filesystem sandbox backed by OS-native primitives (Landlock, Seatbelt, Windows ACLs)",
+    "license": "MIT",
+    "web": "https://github.com/capocasa/sandwall"
   }
 ]


narimiran:
Unfortunately, there is an existing (unmaintained? last commit 7 years ago) package with the same name in the package list: https://github.com/dom96/nimbox

Niminem:
@narimiran it's Dom. Might as well remove that one from the pkg list. That dude's not coming back to Nim maintaining anything, I think he's still crying somewhere in a corner

hugosenari:
> @narimiran it's Dom. Might as well remove that one from the pkg list. That dude's not coming back to Nim maintaining anything, I think he's still crying somewhere in a corner

The issue, is that may (or not) break someone else package that depends on that one.

Niminem:
@hugosenari doubt it enough to remove it if I could. Last commit 8 years ago, was 'in development' at the time of publishing, built with nim 0.14.3 hardly anyone is using refc, and the auto-generated bindings is for termbox via a simple c2nim call (which is likely broken by now)

hugosenari:
> @hugosenari doubt it enough to remove it if I could. Last commit 8 years ago, was 'in development' at the time of publishing, built with nim 0.14.3 hardly anyone is using refc, and the auto-generated bindings is for termbox via a simple c2nim call (which is likely broken by now)

Ok, but we should at least have a doc with rules, and include something like: removed because owner lisked BDfL give a shit about wokeism.
Or less ofensive: Remove because is irrelevant (no know package using it), no update in X years, AND someone else trying to take that 'namespace'.
But think namespace per owner, is a better solution https://github.com/nim-lang/packages/pull/3365#issuecomment-4565916414

Niminem:
@hugosenari namespace per owner definitely the best option

hugosenari:
image

capocasa:
Thank you, @Niminem! Sorry I didn't notice it was taken.

OK I'll leave this PR here for documentation- I'll just come back when I have another name, it's no big deal.

For the bigger picture- I did some [quick research](https://gist.github.com/capocasa/6154efee6b1e19c2684a45c7ee0a3cb5) and it seems handling unmaintained packages varies widlly. Python, Debian and Arch Linux have some kind of orphaning and taking over process, while Go and Rust packages stay forever.

capocasa:
Renamed the lib to sandwall! Should be conflict free now.

PR #3380, opened 2026-06-26
Add "sonaliwan" to the packages.json list.

jihem:
I added "sonaliwan" to the list in the `packages.json` file. I used Zed, which normalized the entries. At first glance, the number of changes is surprising, but it was necessary to clean things up. If it drives you crazy, I'll do it again without changing anything else (so the file keeps evolving in the ugliest way possible 😂). Regards, jihem


narimiran:
> If it drives you crazy, I'll do it again without changing anything else

It does.

PR #3389, opened 2026-06-30
Add package yyjson

zystem:
Thin Nim bindings for yyjson with a small idiomatic high-level API

https://github.com/zystem/nim-yyjson


narimiran:
You've already added this package to the list, no need to do it again.

PR #3417, opened 2026-07-13
Add package gitignore

Niminem:
dependency-free, spec-compliant gitignore parser and pattern matcher

https://github.com/Niminem/gitignore


Niminem:
accidently tried pulling w/ a private repo. updated to https://github.com/nim-lang/packages/pull/3418

PR #3446, opened 2026-08-03
Add package nimanim

himanshm:
A port of ManimCE in Nim (with optional voiceover support)

https://github.com/himanshm/nimanim


narimiran:
You've already added that package.

PR #3463, opened 2026-08-12
rename nimantic_versioning to nimver

vinpogo:
I decided to shorten the name to be less verbose


packages.json
```diff
@@ -1,7 +1,11 @@
 [
-  {
+  {
```
narimiran:
```suggestion
  {
```

PR #3483, opened 2026-08-24
Add terminal_tables

titanomachy:
Pure-Nim responsive terminal tables with rich styling, Unicode/ANSI-aware layouts, and resize-safe live data dashboards.


titanomachy:
Pausing while I rename the package family to term_styles, term_tables, and term_graphs.

PR #3501, opened 2026-08-31
Add drawit package

jihem:
DrawThings(.io) nim client using http api.


narimiran:
> E: Duplicate package drawIt from url https://gitlab.com/OFThomas/drawIt
> E: Duplicate package drawit from url https://github.com/lipunila-sonaliwan-fr/nim.drawit.git

The package with that name already exists. You'll have to pick another one.

